PR #5, opened 2022-09-28
Add genertion of tests results in GitHub Action Summary

Ethyling:
This PR adds a step to generate a test result that will be added to the GitHub Action Summary of the test stages

### DIFF
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -11,6 +11,9 @@ on:
       extended_nightly_tests:
         type: boolean
         default: false
+      tests_result_dir:
+        type: string
+        default: "test-results"
 
 jobs:
   compute-matrix:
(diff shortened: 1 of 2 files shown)


ajschmidt8:
`pytest` generates XML reports as well, so we should see if we can add the summaries to `conda-python-tests.yaml` also.

- https://github.com/rapidsai/rmm/blob/09208b4a9a086db8f2b90984522114bd6deb8e5c/ci/test_python.sh#L39

Ethyling:
> `pytest` generates XML reports as well, so we should see if we can add the summaries to `conda-python-tests.yaml` also.
> 
>     * https://github.com/rapidsai/rmm/blob/09208b4a9a086db8f2b90984522114bd6deb8e5c/ci/test_python.sh#L39

Yes I will, I'm just testing with CPP for now.

ajschmidt8 (approved):

PR #7, opened 2022-10-04
Update workflows to use `sha` and `date`

ajschmidt8:
This PR adds a `sha` and `date` input to our workflows.They are used to ensure that the correct code is built and tested for each nightly run.

Additionally, the existing `ref` input was renamed to `branch` since we generally pass a branch name to that input.


ajschmidt8:
xref https://github.com/rapidsai/actions/pull/1
xref https://github.com/rapidsai/rmm/pull/1123

PR #8, opened 2022-10-04
Fix test matrix

ajschmidt8:
This PR removes the CUDA `11.2.2` `arm` combinations from our tests matrices. `arm` images are only provided for CUDA versions greater than (but not equal to) `11.2.x`.


ajschmidt8:
see https://hub.docker.com/r/rapidsai/ci/tags?page=1&name=cuda11.2.2-ubuntu18.04-py3.9

PR #4, opened 2022-09-23
Add workflow for wheels

sevagh:
This PR adds a workflow for building RAPIDS wheels compliant for PyPI using cibuildwheel on manylinux containers.

The workflow is used in all of these projects to build wheels (mostly successfully, with some remaining challenges):
* RMM wheels: https://github.com/rapidsai/rmm/pull/1074
* cuDF wheels: https://github.com/rapidsai/cudf/pull/11402
* RAFT wheels: https://github.com/rapidsai/raft/pull/765
* cuml wheels: https://github.com/rapidsai/cuml/pull/4870
* cugraph wheels: https://github.com/rapidsai/cugraph/pull/2606

New PRs:
* ptxcompiler wheels: https://github.com/rapidsai/ptxcompiler/pull/27
* cubinlinker wheels: https://gitlab-master.nvidia.com/shanssian/cubinlinker
* RMM wheels: https://github.com/rapidsai/rmm/pull/1107
* ucx-py wheels: https://github.com/rapidsai/ucx-py/pull/881

* RAFT wheels: https://github.com/rapidsai/raft/pull/849 (raft_dask, pylibraft)
* cuDF wheels: https://github.com/rapidsai/cudf/pull/11790 (dask_cudf, cudf)

* cuml wheels: https://github.com/rapidsai/cuml/pull/4905
* cuGraph wheels: https://github.com/rapidsai/cugraph/pull/2747 (pylibcugraph, cugraph)

PR #16, opened 2022-11-03
Add workflow for wheels

sevagh:
This PR replaces PR #4 (so we can preserve the original while we work on both side by side).


.github/workflows/wheels-pure.yml
```diff
@@ -0,0 +1,132 @@
...
+
+    steps:
+    - name: checkout code repo
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0 # unshallow fetch for setuptools-scm
+        persist-credentials: false
+
+    - name: Build pure python wheel with pip wheel
+      run: python-cp38 -m pip wheel -w ./dist ${{ inputs.package-dir }} --no-deps
```
sevagh, line 58: Note: this is my own symlink, maybe think about doing it differently

.github/workflows/wheels-pure.yml
```diff
@@ -0,0 +1,132 @@
...
+        RAPIDS_PY_VERSION: "3.8"
+        AWS_ACCESS_KEY_ID: ${{ secrets.RAPIDSAI_GHA_AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.RAPIDSAI_GHA_AWS_SECRET_ACCESS_KEY }}
+
+  wheel-test:
+    name: wheel test pure
+    runs-on: [self-hosted, linux, amd64, gpu-v100-495-1]
+    needs: wheel-build
+    container:
+      image: "rapidsai/citestwheel_x86_64"
```
sevagh, line 77: Merge cibuildwheel/citestwheel containers (they're mostly the same)

sevagh: It takes ~1 minute to download the runtime cibw launcher, and ~3-4 minutes to download the devel citestwheel tester - keeping them separate is fine.

vyasr (approved):
I'm pretty happy with this workflow at this stage. There are a couple of variables that are probably extraneous right now but we will need later like docker flags, and I wonder if maybe the cuda-suffix shouldn't even be configurable, but I think we're nitpicking now so let's merge and just iterate as we go. 

Thanks for working through all the issues here!

ajschmidt8 (approved):

PR #18, opened 2022-11-16
Make test scripts configurable

ajschmidt8:
This PR makes the CPP and Python test scripts configurable. This enables us to use these shared workflows in a parent workflow multiple times with different shell scripts. The benefit of this is that we can parallelize Python package tests (e.g. testing `cudf`, `dask-cudf`, `custreamz`, etc. at the same time).

### DIFF
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -18,6 +18,9 @@ on:
       tests_result_dir:
         type: string
         default: "test-results"
+      test_script:
+        type: string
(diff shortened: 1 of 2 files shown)


bdice (approved):
Can we do the same for build scripts?

ajschmidt8:
> Can/should we do the same for build scripts?

I can make that change now since it's trivial. I just don't want us to get bogged down with trying to over-optimize parallelization between builds and tests before we've shipped a GH Actions MVP for each repository.

sevagh (approved):

PR #19, opened 2022-11-18
Add custom-job.yaml

bdice:
This PR adds a workflow called `custom-job.yaml` that can be used for an arbitrary job (build, test, ...) that is not a full matrix build like typical C++ or Python builds/tests. This draft PR is being tested out in https://github.com/rapidsai/cudf/pull/12002 for Java tests, which are only run on one arch/version combination and thus can't use the existing C++/Python testing workflows.

### DIFF
--- a/.github/workflows/custom-job.yaml
+++ b/.github/workflows/custom-job.yaml
@@ -1,0 +1,57 @@
+on:
+  workflow_call:
+    inputs:
+      build_type:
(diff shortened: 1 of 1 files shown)


bdice:
Closing this PR from my fork in favor of #20. It needs to come from `rapidsai` in order to be tested.

PR #20, opened 2022-11-18
Add custom-job.yaml, RAPIDS_BASE_BRANCH, Codecov support

bdice:
This PR adds features needed for cuDF GitHub Actions workflows:

- A workflow called `custom-job.yaml` has been added that can be used for an arbitrary job (build, test, ...) that is not a full matrix build like typical C++ or Python builds/tests. This is needed for Java tests and notebook tests, which are only run on one arch/version combination and thus can't use the existing C++/Python testing workflows.

- "Get PR Info" steps inject `RAPIDS_BASE_BRANCH` into the CI environment (needed for copyright style checks).

- Codecov support.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -42,6 +42,10 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - uses: actions/cache@v3
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit-0|${{ hashFiles('.pre-commit-config.yaml') }}
       - name: Run ci/check_style.sh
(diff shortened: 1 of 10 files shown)


ajschmidt8 (approved):

ajschmidt8:
Just want to add another note based on offline discussions. We opted to include `custom-job.yaml` to this repository to keep the environment variable standardization snippet below in the same place as the other workflows which use similar standardization. This way if we need to update that logic in the future, it can all be done in this single repository.

- https://github.com/rapidsai/shared-action-workflows/blob/3546a5fedd6a098321061e1d02243f4d80cbbeba/.github/workflows/custom-job.yaml#L52-L56

PR #23, opened 2022-11-30
Use `codecov` from CI images

ajschmidt8:
This PR includes the following changes:

- Updates the `codecov` implementation to use the binary included in our CI images
  - This is to address the intermittent timeouts that we were experiencing with `codecov/codecov-action@v3`
- Ensures that `codecov` only runs on `x86`/`x86_64` machines
  - codecov doesn't support `arm`, see [this issue](https://github.com/codecov/uploader/issues/523)
- Removes the `build_type == 'pull-request'` restriction from the `codecov` step
  - This is to ensure we have `codecov` results for `base` branches (see [here](https://docs.codecov.com/docs/error-reference#missing-base-report) for why that's important)
- Creates a new `RAPIDS_TESTS_DIR` environment variable whose value is a directory path where unit test results can be saved to
  - This is similar to the `RAPIDS_COVERAGE_DIR` environment variable from #20
  - As a result of this change, I've also removed the `tests_result_dir` input

### DIFF
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -118,18 +118,7 @@ jobs:
       - name: Run codecov
         if: inputs.run_codecov && inputs.build_type == 'pull-request'
         run: |
-          #!/bin/bash
-          set -ex
-          env | sort
-          curl https://uploader.codecov.io/verification.gpg --max-time 10 --retry 5 | gpg --no-default-keyring --keyring trustedkeys.gpg --import
-          curl -Os --max-time 10 --retry 5 https://uploader.codecov.io/latest/linux/codecov
-          curl -Os --max-time 10 --retry 5 https://uploader.codecov.io/latest/linux/codecov.SHA256SUM
-          curl -Os --max-time 10 --retry 5 https://uploader.codecov.io/latest/linux/codecov.SHA256SUM.sig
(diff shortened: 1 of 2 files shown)


.github/workflows/conda-python-tests.yaml
```diff
@@ -113,10 +111,12 @@ jobs:
...
-          paths: "${{ inputs.tests_result_dir }}/*.xml"
+          paths: "${{ env.RAPIDS_TESTS_DIR }}/*.xml"
         if: always()
-      - uses: codecov/codecov-action@v3
-        if: inputs.run_codecov && inputs.build_type == 'pull-request'
-        with:
-          directory: ${{ env.RAPIDS_COVERAGE_DIR }}
-          verbose: true
+      - name: Run codecov
+        if: inputs.run_codecov && contains(fromJson('["X64", "X86"]'), runner.arch)
```
bdice, line 117: We probably only need one value here, right? Which is it?

bdice (approved):

.github/workflows/conda-python-tests.yaml
ajschmidt8: Probably only `X64`. I got the full list of values at the link below. I'll update it.

- https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context

PR #24, opened 2022-12-02
Remove `merge-gatekeeper`

ajschmidt8:
This PR removes `merge-gatekeeper`. We have to remove it due to some rate-limiting errors we were getting on high-volume repositories like `cudf`.

According to the GH docs below, the `GITHUB_TOKEN` provided by GitHub Actions is limited to 1000 requests per repository per hour. This limit is not enough for repos like `cudf`.

https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#requests-from-github-actions

### DIFF
--- a/.github/workflows/pr-builder.yaml
+++ b/.github/workflows/pr-builder.yaml
@@ -5,11 +5,4 @@ jobs:
   run:
     runs-on: ubuntu-latest
     steps:
-      - uses: upsidr/merge-gatekeeper@v1
-        with:
-          self: "pr-builder / run"
-          token: ${{ secrets.GITHUB_TOKEN }}
-          interval: 60
-          timeout: 86400 # 24h
-          ref: ${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-builder.yaml
```diff
@@ -5,11 +5,4 @@ jobs:
...
     steps:
-      - uses: upsidr/merge-gatekeeper@v1
-        with:
-          self: "pr-builder / run"
-          token: ${{ secrets.GITHUB_TOKEN }}
-          interval: 60
-          timeout: 86400 # 24h
-          ref: ${{ github.ref }}
-          ignored: "triage,Branch Checker,Label Checker"
+      - run: exit 0
```
bdice, line 8: Maybe add a comment to explain how this is supposed to work.
```suggestion
      # This workflow should depend on all other workflows in a repository using "needs:"
      - run: exit 0
```

bdice (approved):

PR #27, opened 2022-12-07
Upkeep/manylinux improvements

sevagh:
**Introduces breaking changes to current wheels.yml files, don't merge yet**

I'm opening this PR to contain the various improvements to the manylinux/cibuildwheel workflow planned for 23.02

This was actually used at the tail end of 22.12 to help build ptxcompiler and cubinlinker with CTK 11.8


.github/workflows/wheels-manylinux-publish.yml
```diff
@@ -0,0 +1,68 @@
...
+        sha="${{ inputs.sha }}"
+        if [[ "${sha}" == "" ]]; then
+          sha=$(git rev-parse HEAD)
+        fi
+        echo "RAPIDS_SHA=${sha}" | tee --append "${GITHUB_ENV}" "${GITHUB_OUTPUT}"
+        echo "RAPIDS_REF_NAME=${{ inputs.branch || github.ref_name }}" | tee --append "${GITHUB_ENV}" "${GITHUB_OUTPUT}"
+        echo "RAPIDS_NIGHTLY_DATE=${{ inputs.date }}" | tee --append "${GITHUB_ENV}" "${GITHUB_OUTPUT}"
+
+    - name: Download wheels from downloads.rapids.ai and publish to internal PyPI
+      run: |
```
vyasr, line 59: Should we implement [a tool like `rapids-upload-to-anaconda`](https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-upload-to-anaconda)? Maintaining the matrix in the form of a bunch of repeated lines like this isn't great and would be better suited to a script where we have true loop constructs available to handle that.

.github/workflows/wheels-manylinux-build.yml
```diff
@@ -0,0 +1,201 @@
...
+    steps:
+    - name: checkout code repo
+      uses: actions/checkout@v3
+      with:
+        repository: ${{ inputs.repo }}
+        ref: ${{ inputs.sha }}
+        fetch-depth: 0 # unshallow fetch for setuptools-scm
+        persist-credentials: false
+
+    - name: Standardize repository information
```
vyasr, line 128: This step seems identical in every wheel workflow, and it's also present in the [conda workflows](https://github.com/rapidsai/shared-action-workflows/blob/main/.github/workflows/conda-upload-packages.yaml). Can we extract it into a composite action or a reusable workflow (I don't recall off the top of my head which -- if either -- would be suitable for things like the `GITHUB_ENV`/`GITHUB_OUTPUT` modifications done in this step).

.github/workflows/wheels-manylinux-build.yml
```diff
@@ -0,0 +1,201 @@
...
+      uses: actions/checkout@v3
+      if: ${{ inputs.extra-repo != '' }}
+      with:
+        repository: ${{ inputs.extra-repo }}
+        ref: ${{ inputs.extra-repo-sha }}
+        path: ${{ steps.preprocess-extras.outputs.EXTRA_REPO_PATH }}
+        ssh-key: ${{ secrets[inputs.extra-repo-deploy-key] }}
+        persist-credentials: false
+
+    - name: Set CTK-related vars from input CTK versions
```
vyasr, line 156: Another step that could maybe be shared across workflows. I see that the wheel test only uses the wheel name and not the cuda suffix, so could make sense to split that up, although it might also be fine to just have the extra var floating around in that case.

.github/workflows/wheels-manylinux-test.yml
```diff
@@ -0,0 +1,165 @@
...
+      # general settings
+      package-name:
+        required: true
+        type: string
+
+      # test settings
+      test-docker-options:
+        required: false
+        type: string
+        default: '-e _NOOP'
```
vyasr, line 28: I don't recall exactly how ucx-py uses this option, I assume something breaks if the default here is removed altogether.

vyasr (commented):
This is looking really good! I made some comments as I went through to familiarize myself with the new layout, but I realize that this PR is still a draft and you're probably well aware of the issues that I pointed out so feel free to ignore any part of the review.

.github/workflows/wheels-manylinux-publish.yml
sevagh: My thoughts on this are mixed;

The matrix for building (pyver, arch) is decided in this workflow. The matrix for publishing corresponds to the matrix for building.

E.g. to change py 3.9 to 3.10 would be a sed command of 's/3.9/3.10/g' in the wheel workflow files in this repo, without involving gha-tools.

.github/workflows/wheels-manylinux-build.yml
sevagh: Yeah I'll see if that's possible (without creating some unwanted behavior such as the modification of GITHUB_* depending on whether the step is in a composite action vs. embedded in a reusable workflow).

.github/workflows/wheels-manylinux-build.yml
sevagh: Actually, in setting the stage to support CTK 12 in the future, I've gone around appending `_cu11` (or `_cu${ctk_major}`) to the `package-name` i.e. `RAPIDS_PY_WHEEL_NAME`

Old (main branch) workflow:
* RMM: rmm_wheel_python_rmm_x86_64_3.8.tar.gz

New (this WIP PR) workflow, before CTK 12 is added:
* RMM: rmm_wheel_python_rmm_cu11_x86_64_3.8.tar.gz

Future (when CTK 12 is added):
* RMM: rmm_wheel_python_rmm_{cu11, cu12}_x86_64_3.8.tar.gz

.github/workflows/wheels-manylinux-build.yml
sevagh: Seems to have worked fine

.github/workflows/wheels-manylinux-publish.yml
sevagh: I just used GHA's built-in matrix again. This way the same matrix reflects across build, test, publish.

.github/workflows/wheels-manylinux-build.yml
```diff
@@ -0,0 +1,192 @@
...
+      - self-hosted
+      - linux
+      - ${{ matrix.arch }}
+      - cpu16
+    container:
+      # ctk version of the cibw container is irrelevant in the manylinux case
+      # it's simply a launcher for cibuildwheel
+      image: "rapidsai/cibuildwheel:cuda-runtime-11.8.0-ubuntu20.04"
+      volumes:
+        - /var/run/docker.sock:/var/run/docker.sock
```
vyasr, line 112: I assume this will always be needed for the docker in docker with manylinux? May be worth noting here.

.github/workflows/wheels-manylinux-publish.yml
```diff
@@ -0,0 +1,71 @@
...
+      # it's simply a launcher for twine
+      image: "rapidsai/cibuildwheel:cuda-runtime-11.8.0-ubuntu20.04"
+      env:
+        RAPIDS_PY_WHEEL_BUILD_TAG: ${{ inputs.build-tag }}
+        RAPIDS_BUILD_TYPE: ${{ inputs.build_type }}
+        RAPIDS_ARCH: ${{ matrix.linux-arch }}
+        RAPIDS_PY_VERSION: ${{ matrix.python }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.RAPIDSAI_GHA_AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.RAPIDSAI_GHA_AWS_SECRET_ACCESS_KEY }}
+        TWINE_USERNAME: cibuildwheel
```
vyasr, line 50: Out of scope for now, but I would probably rename our user on k8s to rapids or something rather than cibuildwheel.

.github/workflows/wheels-manylinux-test.yml
```diff
@@ -0,0 +1,145 @@
...
+      BUILD_TYPE: ${{ inputs.build_type }}
+    outputs:
+      MATRIX: ${{ steps.compute-matrix.outputs.MATRIX }}
+    steps:
+      - name: Validate test type
+        run: |
+          if [[ "$BUILD_TYPE" != "pull-request" ]] && [[ "$BUILD_TYPE" != "nightly" ]]; then
+              echo "Invalid build type! Must be 'nightly' or 'pull-request'."
+              exit 1
+          fi
```
vyasr, line 63: Do we need this step? This seems like the equivalent of an assert that you had in dev but can be dropped at this point unless I'm missing something.

rapids-github-info/action.yml
```diff
@@ -0,0 +1,35 @@
...
+    value: ${{ steps.standardize-repo-info.outputs.RAPIDS_REPOSITORY }}
+  rapids_sha:
+    value: ${{ steps.standardize-repo-info.outputs.RAPIDS_SHA }}
+  rapids_ref_name:
+    value: ${{ steps.standardize-repo-info.outputs.RAPIDS_REF_NAME }}
+  rapids_nightly_date:
+    value: ${{ steps.standardize-repo-info.outputs.RAPIDS_NIGHTLY_DATE }}
+
+runs:
+  using: "composite"
```
vyasr, line 22: Don't think the quotes are necessary here.

vyasr (approved):
This looks great! There's still a lot of duplication in the form of GHA boilerplate for inputs and such, but I don't think there's a whole lot that we can do about that right now. Getting these stages split out to support alignment with conda is huge.

.github/workflows/wheels-manylinux-test.yml
sevagh: I just copied it from Conda GHA

.github/workflows/wheels-manylinux-test.yml
sevagh: Yes, we have to use the `options:` key of the `container:` block of GHA (line #99 of this file), and `-e _NOOP` just passes a dummy env var called _NOOP to have a valid string. Empty strings lead to a `docker ... <args> ''` which is invalid.

PR #29, opened 2022-12-13
Only run shared workflows on rapidsai repositories.

### LINKED ISSUE #28: Forks are attempting to run GitHub Actions
Looking at my fork of cudf, I see that it tries to run GitHub Actions. https://github.com/bdice/cudf/actions
Example run (failure): https://github.com/bdice/cudf/actions/runs/3681431324
It fails with the message:
```
Called workflows cannot be queued onto self-hosted runners across organisations/enterprises. Failed to queue this job. Labels: 'self-hosted , linux , cpu8 , amd64'.
```

We may need to "protect" actions that use self-hosted runners so that they don't trigger on forks. One way to do this is to add a conditional to the job definition:

```yaml
jobs:
  build:
    if: ${{ github.repository_owner == 'rapidsai' }}
```

bdice:
Closes #28.

@ajschmidt8 Should this apply to style checks too? If so, it'd be nice to have your eyes on the syntax.

### DIFF
--- a/.github/workflows/conda-cpp-matrix-build.yaml
+++ b/.github/workflows/conda-cpp-matrix-build.yaml
@@ -21,6 +21,7 @@ on:
 
 jobs:
   build:
+    if: ${{ github.repository_owner == 'rapidsai' }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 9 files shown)


bdice:
Okay. For now I disabled GitHub Actions on my fork. I think we'll need to point a lot of folks in that direction because otherwise pushing to your fork will send you an email about the failed workflow, for every repo.

PR #30, opened 2022-12-19
Use updated image names for wheel workflows

sevagh:
The wheels workflow were using some bad image names:
* Having the architecture in the image name (or tag name), vs. using platform/arch tag features of docker to use the native architecture of the runner host
* Using an unspecified implicit `:latest` instead of an explicit tag
* Lack of detail of the image (which OS, which CUDA version, etc.)

I have since overhauled the naming scheme for wheel-related images; for example, `cibuildwheel_x86_64:latest` is replaced with its more informative new name, `cibuildwheel:cuda-runtime-11.5.1-ubuntu18.04`.

It was tested to build cuDF (testing both the pure and manylinux workflows):
https://github.com/rapidsai/pypi-wheel-scripts/actions/runs/3733934876 
(this is a custom cuDF build job that points to this branch)

This is a small PR to ensure that existing wheels using `@main` don't break when I clean up the old invalid image names. 

The goal is to cleanly support CUDA 11 and 12 in 23.02, by having the CTK version in the image name. There will be a bigger 23.02 PR which will encompass bigger wheel workflow changes.

### DIFF
--- a/.github/workflows/wheels-manylinux.yml
+++ b/.github/workflows/wheels-manylinux.yml
@@ -57,11 +57,11 @@ on:
       manylinux-container-amd64:
         required: false
         type: string
-        default: "rapidsai/manylinux2014_x86_64"
+        default: "rapidsai/manylinux2014:cuda-devel-11.5.1-centos7"
       manylinux-container-arm64:
         required: false
         type: string
-        default: "rapidsai/manylinux_2_31_aarch64"
+        default: "rapidsai/manylinux_2_31:cuda-devel-11.5.1-ubuntu20.04"
       auditwheel-repair-command-amd64: # hint: use 'cp {wheel} {dest_dir}' to skip repairing
         required: false
         type: string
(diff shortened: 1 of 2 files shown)


ajschmidt8 (approved):
I see some discrepancies in the Ubuntu version used.

e.g. `rapidsai/cibuildwheel:cuda-runtime-11.5.1-ubuntu18.04` vs. `rapidsai/cibuildwheel:cuda-runtime-11.5.1-ubuntu20.04`.

Not sure if that's intentional or not, but figured I'd call it out. Otherwise, LGTM!

sevagh:
These are intentional discrepancies, related to the manylinux environments where we build wheels, and how it influences where we can test wheels:

* x86_64 wheels are built on CentOS 7 and tested on Ubuntu 18.04 (libc 2.17 for building, libc 2.27 for testing)
* aarch64 wheels are built on Ubuntu 20.04 and tested on Ubuntu 20.04 (libc 2.31 for building and testing)

The total effect is a patchwork of different images available on different architectures, summarized in the cibuildwheel-imgs README:
<https://github.com/rapidsai/cibuildwheel-imgs#cibuildwheel-imgs>

PR #31, opened 2022-12-31
Add shared job to check `pr-builder` job dependencies

ajschmidt8:
This PR adds a new check to run `rapids-check-pr-job-dependencies`.

Details on what that script does and why it's needed can be found below.

- https://github.com/rapidsai/gha-tools/pull/30

Additionally, the PR consolidates some checks that aren't likely to fail very often (e.g. `size-checker` and this new `pr-job-dependencies` check) into a new job called `other-checks`.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -24,8 +24,7 @@ on:
         required: false
 
 jobs:
-  check-size:
-    if: ${{ inputs.enable_check_size }}
+  other-checks:
     runs-on: ubuntu-latest
     container:
       image: rapidsai/ci:latest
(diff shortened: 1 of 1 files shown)


bdice (approved):

ajschmidt8:
I've tested this script locally, but I want to test it in a PR before we merge it.

ajschmidt8:
I tested these changes successfully in https://github.com/rapidsai/rmm/pull/1185.

I've consolidated the `size-checker` and this new `pr-job-dependencies` check into a new `other-checks` job.

This will ensure that the existing number of checks on PRs doesn't grow as a result of these changes.

I think there's value in having `check-style` and `check-generated-files` as their own jobs since it lets users see if the status of those checks directly from the PR page (without having to click into the GH Action job for more details).

For checks that aren't likely to fail often (e.g. `size-checker`, `pr-job-dependencies`), I think it's fine to consolidate them into this new `other-checks` job.

@bdice, thoughts?

bdice:
Good enough for me! I would have consolidated the dependencies generated files as well, since it only takes a couple clicks to see the failing step. Either way is fine.

PR #41, opened 2023-02-07
Add & run `update-version.sh`, Remove Python `3.9` builds

ajschmidt8:
This PR includes the following changes:

- Adds an `update-version.sh` script
- Runs the `update-version.sh` script in order to update all branch references to `branch-23.02`
- Removes Python `3.9` builds from `branch-23.02` since RAPIDS no longer builds that version
- Updates conda C++ builds to use latest Ubuntu / Python images
- Updates conda Python builds to use latest Ubuntu images
- Removes ubuntu 18.04 from tests

The implication of the second bullet is that every time we create a new branch, all of the composite action references will also be updated.

I think that's fine, but it's something we'll need to be conscious of.

I think we should try to keep the reusable workflow branches and composite action branches in-sync.

For example, if we need to test a composite action change, we should do so on a separate branch. And then update the reusable workflow to call the newly changed composite action on that branch.

Thoughts on this, @sevagh?

Keeping these workflows and composite actions in-sync will become easier if/when the following discussions are ever addressed by GitHub. I'm currently subscribed to them.

- https://github.com/community/community/discussions/18601
- https://github.com/community/community/discussions/18602

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -35,7 +35,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Get PR Info
         id: get-pr-info
-        uses: rapidsai/shared-action-workflows/get-pr-info@main
+        uses: rapidsai/shared-action-workflows/get-pr-info@branch-23.02
       - name: Run rapids-size-checker
         if: ${{ inputs.enable_check_size }}
         run: rapids-size-checker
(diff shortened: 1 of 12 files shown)


raydouglass (approved):

sevagh (approved):

vyasr:
> For example, if we need to test a composite action change, we should do so on a separate branch. And then update the reusable workflow to call the newly changed composite action on that branch.

Is this equivalent to saying that a reusable workflow should never point to a composite action on another branch? I would be in favor of that, I think keeping those separate is much more stable and less error-prone.

ajschmidt8:
> Is this equivalent to saying that a reusable workflow should never point to a composite action on another branch? I would be in favor of that, I think keeping those separate is much more stable and less error-prone.

Yes, that's a better way of phrasing it.

raydouglass (approved):

PR #43, opened 2023-02-14
Use AWS OIDC to get AWS creds

jjacobelli:
This PR is using the AWS OIDC to get AWS credentials ([doc](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services)). The `AWS_ROLE_ARN` and `AWS_REGION` are orgs variables defined here: https://github.com/organizations/rapidsai/settings/variables/actions

### DIFF
--- a/.github/workflows/wheels-manylinux-build.yml
+++ b/.github/workflows/wheels-manylinux-build.yml
@@ -106,6 +106,8 @@ jobs:
         python: ['3.8', '3.10']
         arch: [amd64, arm64]
         ctk: ['11.8.0']
+    permissions:
+      id-token: write
     runs-on:
       - self-hosted
       - linux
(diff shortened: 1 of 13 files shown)


ajschmidt8:
Let's use the `v1-node16` branch to avoid the `node12` deprecation warning as described below.

- https://github.com/aws-actions/configure-aws-credentials/tree/1416c559c08e8051b3ff6e0739e73ac4bb0d6c57#notice-node12-deprecation-warning

jjacobelli:
PR used to test on `rmm`: https://github.com/rapidsai/rmm/pull/1210

.github/workflows/wheels-manylinux-build.yml
```diff
@@ -208,8 +212,8 @@ jobs:
         uses-setup-env-vars: ${{ inputs.uses-setup-env-vars }}
 
         # secrets
-        sccache-aws-access-key-id: ${{ secrets.RAPIDSAI_GHA_AWS_ACCESS_KEY_ID }}
-        sccache-aws-secret-access-key: ${{ secrets.RAPIDSAI_GHA_AWS_SECRET_ACCESS_KEY }}
+        sccache-aws-access-key-id: ${{ env.AWS_ACCESS_KEY_ID }}
+        sccache-aws-secret-access-key: ${{ env.AWS_SECRET_ACCESS_KEY }}
```
ajschmidt8, line 216: we probably also want to pass in the value of the new `AWS_SESSION_TOKEN` variable here.

Here are the AWS docs that talk about temporary credentials and the session token for reference.

- https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html


CCing @sevagh for awareness about this change (no action is required).

PR #38, opened 2023-01-14
Replace versioneer overrides with manual modification of package source

vyasr:
This PR replaces setting the `RAPIDS_PY_WHEEL_VERSIONEER_OVERRIDE` environment variable for consumption within Python packages with instead directly modifying metadata in the package. It relies on each package to provide a script `ci/release/modify_wheel_build_version.sh` that can be executed at wheel build time to perform all the necessary version replacements.

PR #45, opened 2023-02-16
Allow scripts to overwrite package names with a prerun script rather than using an env variable

vyasr:
This PR builds on #38 to remove environment variable-based dynamic names from wheels, instead using the same script defined by #38 to modify any static files directly.

### DIFF
--- a/.github/workflows/wheels-pure-build.yml
+++ b/.github/workflows/wheels-pure-build.yml
@@ -29,16 +29,12 @@ on:
         required: false
         default: 'true'
 
-      # TODO: The below flags are to support the older packaging structure. We can
-      # remove them once all packages building wheels have migrated away from the
-      # old structure.
-      # Enable libraries that aren't using versioneer
-      uses-versioneer:
-        required: false
-        type: boolean
-        default: true
-      # Enable libraries that don't use RAPIDS_PY_WHEEL_CUDA_SUFFIX in setup.py
-      uses-dynamic-name:
+      # TODO: The below flag is to support the older packaging structure. We
+      # can remove it once all packages building wheels have migrated away from
+      # the old structure.
+      # Enable libraries that aren't using versioneer or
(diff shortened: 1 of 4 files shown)


.github/workflows/wheels-pure-build.yml
```diff
@@ -94,9 +106,11 @@ jobs:
 
         if [ "${{ inputs.uses-versioneer }}" = true ]; then
           export RAPIDS_PY_WHEEL_VERSIONEER_OVERRIDE="${versioneer_override}"
-        else
-          bash ci/release/apply_wheel_modifications.sh ${versioneer_override}
-          echo "The git diff after modifying the versions is:"
+        fi
+
+        if [ "${{ inputs.uses-versioneer }}" = false ] || [ "${{ inputs.uses-dynamic-name }}" = false ]; then
```
sevagh, line 111: Just checking; is equality in bash a single equals sign? Or is that for single brackets?

vyasr: Lol this is one of those things I hate, a lot of these things depend on which shell and what you have. I believe single equals is equality and is standard (i.e. works across different shell types). For double equals the behavior is shell-dependent, but in bash it would give you the same result for booleans. And yes, this is in single brackets; double brackets are also an extension of `sh` but they work on all commonly used shells today.

We could rewrite this statement in a lot of different equivalent ways, I just picked one that worked :shrug:

vyasr:
And for examples,
- Here's an rmm PR demonstrating how this works: https://github.com/rapidsai/rmm/pull/1212
- Here's a cudf PR showing that the new feature really is optional and existing scripts will "just work":  https://github.com/rapidsai/cudf/pull/12798. Note that rmm already has the first part of the migration done (versioneer removal), and since we decided to bundle the changes going forward I can't use rmm as the test repo here.

sevagh (approved):

PR #49, opened 2023-02-23
Build CUDA 12 pip wheels alongside CUDA 11

sevagh:
This adds CUDA 12 to the pip wheel workflow in the GHA "cuda-120" branch.


.github/workflows/wheels-manylinux-test.yml
```diff
@@ -66,13 +66,19 @@ jobs:
           {
             "pull-request": [
               { "arch": "amd64", "python": "3.8", "ctk": "11.8.0", "image": "ubuntu18.04", "test-type": "unit", "test-command": "${{ inputs.test-unittest }}", "gpu": "v100", "driver": "520"  },
-              { "arch": "arm64", "python": "3.8", "ctk": "11.8.0", "image": "ubuntu20.04", "test-type": "smoke", "test-command": "${{ inputs.test-smoketest }}", "gpu": "a100", "driver": "520"  }
+              { "arch": "arm64", "python": "3.8", "ctk": "11.8.0", "image": "ubuntu20.04", "test-type": "smoke", "test-command": "${{ inputs.test-smoketest }}", "gpu": "a100", "driver": "525"  },
```
vyasr, line 69: Was bumping some of the 11.8 runs to driver 525 intentional?

sevagh: Yes, ARM instances are limited so we will only have 525 runners.

.github/workflows/wheels-pure-build.yml
```diff
@@ -56,10 +56,10 @@ jobs:
     needs: wheel-epoch-timestamp
     strategy:
       matrix:
-        ctk: ["11.8.0"]
+        ctk: ["11.8.0", "12.0.1"]
     runs-on: ubuntu-latest
     container:
-      image: "rapidsai/cibuildwheel:cuda-runtime-${{ matrix.ctk }}-ubuntu20.04"
+      image: "rapidsai/cibuildwheel:cuda-runtime-11.8.0-ubuntu20.04"
```
ajschmidt8, line 62: Was hardcoding this to `11.8.0` intentional?

sevagh: Yes, I wanted to test to make sure that the cibuildwheel launcher container was essentially CTK-agnostic.

It's only used on CPU instances for 3 things:
1. Building manylinux wheels by running cibuildwheel (where the important CTK parts are in the second manylinux container, which have 11.8.0 and 12.0.1 variants)
2. Building pure-python wheels (that are tied to a manylinux wheel) i.e. dask-cudf-cu12 which is associated with cudf-cu12; this is done on a cpu instance and the CTK doesn't really matter
3. Running twine upload to publish wheels

Something @vyasr brought up is that the cibuildwheel launcher container doesn't need to have CUDA at all; so we could theoretically easily use an upstream basic Ubuntu container.

However, when I came up with the containers in the cibuildwheel-imgs repo, I thought that using `nvidia/cuda` containers as a base for all 3 images (cibuildwheel launcher, citestwheel, manylinux) would be best for consistence, so I finally settled for a single CUDA runtime container.

ajschmidt8 (approved):

.github/workflows/wheels-pure-build.yml
vyasr: We agreed that any attempt to move away from nvidia containers to a lighter upstream container is a super low priority and not even worth considering until we complete a lot of other higher priority tasks at this point.

.github/workflows/wheels-manylinux-test.yml
vyasr: Got it.

PR #52, opened 2023-03-07
Fix shared workflow permissions

ajschmidt8:
After merging #43, there were some private repositories that were having issues checking out source code in GHAs.

Further investigation revealed that by omitting the other keys in the `permissions` block that was added, we were effectively setting them to `none` ([src](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview)).

To remedy this, this PR includes the following changes:

- Keeps the `id-token` permissions as `write`, but set the remaining permissions as described in the `restricted` column [here](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)
  - We can always adjust this if needed, but I think this is fine
- Moves the `permissions` block to the top of each workflow so that it's applied to all jobs in each workflow file

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -19,6 +19,21 @@ on:
         type: string
         default: "ci/build_cpp.sh"
 
+permissions:
+  actions: none
+  checks: none
+  contents: read
+  deployments: none
+  discussions: none
+  id-token: write
+  issues: none
+  packages: read
+  pages: none
+  pull-requests: none
+  repository-projects: none
+  security-events: none
+  statuses: none
+
 jobs:
   build:
     timeout-minutes: 480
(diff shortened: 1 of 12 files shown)


bdice (approved):
Do we need any “pull-request” permissions to check any metadata on a given PR? Wondering if that matters for our “copy PR” workflows.

Also why do we need “packages: read”?

ajschmidt8:
> Do we need any “pull-request” permissions to check any metadata on a given PR? Wondering if that matters for our “copy PR” workflows.

Nope, that's handled separately in the ops-bot. This won't affect it.


> Also why do we need “packages: read”?

I just used the default values listed under the `restrictive` column here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

bdice:
Are we missing keys like “metadata”?

ajschmidt8:
> Are we missing keys like “metadata”? edit: seems like all other keys are listed.

see this excerpt ([src](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token)):

> You can use the `permissions` key in your workflow file to modify permissions for the `GITHUB_TOKEN` for an entire workflow or for individual jobs. This allows you to configure the minimum required permissions for a workflow or job. When the `permissions` key is used, all unspecified permissions are set to no access, **with the exception of the `metadata` scope, which always gets read access.**


`metadata` was also omitted in the example below, which is why I omitted it.

- https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview

bdice:
Thanks for clarifying. I missed that in the docs.

PR #53, opened 2023-03-08
Set `pull-requests` permissions to `read`

ajschmidt8:
This is a follow-up to #52.

In hindsight, we should set the `pull-requests` permission to `read` like @bdice suggested.

This will ensure that the [get-pr-info](https://github.com/rapidsai/shared-action-workflows/blob/93a3e440be17665911049908058c14c9e8a25304/get-pr-info/action.yml) composite action can use the token for retrieving pull-request metadata.

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -29,7 +29,7 @@ permissions:
   issues: none
   packages: read
   pages: none
-  pull-requests: none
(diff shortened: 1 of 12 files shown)


bdice (approved):
[I won't say I told you so. 😉](https://github.com/rapidsai/shared-action-workflows/pull/52#pullrequestreview-1329743097)

PR #57, opened 2023-03-20
Change the default behavior for rewriting wheel names and versions.

vyasr:
All RAPIDS libraries that build wheels have now been migrated away from versioneer and towards the new approach for versioning. In order to avoid any disruptive changes to other repos given the approach of burn down, this PR changes the default value of this parameter rather than removing it outright. That way, once this is merged each library can go through and remove the (now redundant) parameter from their workflows without any breakage, and we can then remove the associated code paths in the workflows in this repo.

### DIFF
--- a/.github/workflows/wheels-manylinux-build.yml
+++ b/.github/workflows/wheels-manylinux-build.yml
@@ -94,7 +94,7 @@ on:
       uses-setup-env-vars:
         required: false
         type: boolean
-        default: true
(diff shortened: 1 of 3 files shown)


vyasr:
Now that we're in burndown for some packages let's push this change to 23.06 to avoid unnecessary breaks.

ajschmidt8 (approved):

PR #62, opened 2023-04-07
Add support for manipulating `conda` build/test matrices downstream

ajschmidt8:
This PR introduces support for manipulating the default matrices in our shared build and test `conda` workflows from downstream repositories.

## Usage:

The filtering can be done via a new workflow input, `matrix_filter`.

`matrix_filter` is intended to contain a `jq` filter (see [docs](https://stedolan.github.io/jq/manual/)).

The `input` provided to `matrix_filter` is an array containing the default build matrix (e.g. [these values](https://github.com/rapidsai/shared-action-workflows/blob/d2eba320e08eedc5087596e47fca89d606058e8c/.github/workflows/conda-cpp-build.yaml#L56-L57)).

The default value of `matrix_filter` is `.` (the `jq` [identity filter](https://stedolan.github.io/jq/manual/#Identity:.)).

This example shows how `matrix_filter` can be used to remove `arm64` matrix combinations from the shared `conda-cpp-build.yaml` workflow.

```yaml
conda-cpp-build:
    uses: rapidsai/shared-action-workflows/.github/workflows/conda-cpp-build.yaml@branch-23.04
    with:
      matrix_filter: map(select(.ARCH != "arm64"))
```

## Additional Changes

This PR also:

- changes the syntax used to represent our default matrices from JSON to YAML since YAML is a little more forgiving. It uses `yq` to convert it back to JSON so that it can be consumed correctly by GHAs.
- switches our generated matrix to look like `{"include": [...]}` instead of `{"includes": [...]}`. This was a typo. Fixing the typo lets us reference the matrix values as `matrix.ARCH` instead of `matrix.includes.ARCH` (see [here](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations) for GHAs docs).

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -59,7 +59,7 @@ jobs:
 
           echo "MATRIX=$(
             yq -n -o json 'env(MATRIX)' | \
-            jq -c '${{ inputs.matrix_filter }} | {includes: .}' \
+            jq -c '${{ inputs.matrix_filter }} | {include: .}' \
           )" >> ${GITHUB_OUTPUT}
   build:
     needs: compute-matrix
@@ -71,11 +71,11 @@ jobs:
       - self-hosted
       - linux
       - ${{ inputs.node_type }}
-      - ${{ matrix.includes.ARCH }}
+      - ${{ matrix.ARCH }}
     env:
       RAPIDS_ARTIFACTS_DIR: ${{ github.workspace }}/artifacts
     container:
-      image: rapidsai/ci:cuda${{ matrix.includes.CUDA_VER }}-${{ matrix.includes.LINUX_VER }}-py${{ matrix.includes.PY_VER }}
+      image: rapidsai/ci:cuda${{ matrix.CUDA_VER }}-${{ matrix.LINUX_VER }}-py${{ matrix.PY_VER }}
       env:
         RAPIDS_BUILD_TYPE: ${{ inputs.build_type }}
         PARALLEL_LEVEL: ${{ env.PARALLEL_LEVEL }}
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -18,6 +18,9 @@ on:
       build_script:
         type: string
         default: "ci/build_cpp.sh"
+      matrix_filter:
+        type: string
+        default: "."
 
 defaults:
   run:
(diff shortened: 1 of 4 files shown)


ajschmidt8:
I assume we'll probably add this to the `wheel` workflows as well, but I want to chat with @sevagh about that first before adding it to this PR.

.github/workflows/conda-cpp-build.yaml
```diff
@@ -39,23 +42,40 @@ permissions:
...
       - self-hosted
       - linux
       - ${{ inputs.node_type }}
-      - ${{ matrix.ARCH }}
+      - ${{ matrix.includes.ARCH }}
     env:
       RAPIDS_ARTIFACTS_DIR: ${{ github.workspace }}/artifacts
     container:
-      image: rapidsai/ci:cuda${{ matrix.CUDA_VER }}-${{ matrix.LINUX_VER }}-py${{ matrix.PY_VER }}
+      image: rapidsai/ci:cuda${{ matrix.includes.CUDA_VER }}-${{ matrix.includes.LINUX_VER }}-py${{ matrix.includes.PY_VER }}
```
bdice, line 78: Is it possible to remove the "includes" level of key here? I am not super familiar with `jq` syntax -- it looks like this is a level of hierarchy that is being added but isn't strictly necessary?

ajschmidt8: Yes, you're right. I battled with this in the past and couldn't figure out why it was necessary for dynamic matrices but not hardcoded matrices.

After re-reading the GHAs docs below, I'm now realizing that my matrix output should say `include` and not `includes`. I think once I fix that typo in the first job, we can remove this extra level in the subsequent job.

- https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations

I'll push a follow-up commit to test.

ajschmidt8:
To Do:

- adapt `wheel` jobs
  - also apply `include` fix to `wheel` jobs
- add `include` changes to PR body

.github/workflows/conda-cpp-build.yaml
ajschmidt8: Yup, that worked in the downstream https://github.com/rapidsai/ucxx/pull/18 PR.

Good catch @bdice!

bdice (approved):
All looks good to me!

ajschmidt8:
`wheel` PR - https://github.com/rapidsai/shared-action-workflows/pull/65

PR #67, opened 2023-04-11
`wheels-manylinux-test.yml` Fix

### LINKED ISSUE #63: Add support for manipulating `wheel` build/test matrices downstream
We should adapt the changes in #62 to `wheel` workflows for consistency.

ajschmidt8:
#63 included a bug that caused the wheel unit and smoke test commands to not be properly escaped.

The fix was to inject those values as `bash` variables rather than with a GitHub Action's expression.

This required updating the outer quotations for the `MATRICES` environment variable in `.github/workflows/wheels-manylinux-test.yml` to use `"` instead of `'`.

For consistency, I've updated the other matrix environment variables to match this style.

Additionally, I've switched the final matrix output to use `tee --append` instead of `>>` so that the matrix value is printed to the GitHub Actions console. This is useful for inspecting the final computed matrices.

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -60,7 +60,7 @@ jobs:
           echo "MATRIX=$(
             yq -n -o json 'env(MATRIX)' | \
             jq -c '${{ inputs.matrix_filter }} | {include: .}' \
-          )" >> ${GITHUB_OUTPUT}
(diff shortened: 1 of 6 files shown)


ajschmidt8:
These workflow changes were tested in https://github.com/rapidsai/cuml/pull/5353.

jjacobelli (approved):

PR #73, opened 2023-05-03
CUDA 12.0 pip support

### LINKED ISSUE #76: Reenable download of locally built wheels at the build stage
The changes in #71 left out the `PIP_FIND_LINKS` logic. The corresponding changes to all of the repos also stopped downloading files to a local wheelhouse in favor of just downloading to a local repo and installing the package in something like a `before-wheel` command. The problem with this change is that it only works for runtime dependencies. Build time dependencies pulled into isolated builds via pyproject.toml build-system:requires lists cannot access this. At present, the only place that this could be a problem is if pylibcugraph's Cython code makes an API-breaking change that needs to be made available to cugraph, because that is the only RAPIDS project building two wheels where the first one is a build-time dependency of the second (pylibraft is a runtime, not build-time, dependency of raft-dask). I think we can afford to wait on getting this fixed until we refactor the workflows to support easier local CI, but if cugraph runs into issues in the interim we can figure out how to get this feature back again.

bdice:
This PR enables CUDA 12.0 CI workflows for pip wheels. It is currently a subset of the changes in the `cuda-120` branch, which also adds conda.

Closes #76

### DIFF
--- a/.github/workflows/wheels-manylinux-test.yml
+++ b/.github/workflows/wheels-manylinux-test.yml
@@ -90,12 +90,18 @@ jobs:
           export MATRICES="
             pull-request:
               - { arch: 'amd64', python: '3.9', ctk: '11.8.0', image: 'ubuntu18.04', test-type: 'unit', test-command: ${UNIT_TEST_CMD}, gpu: 'v100', driver: 'latest' }
+              - { arch: 'amd64', python: '3.9', ctk: '12.0.1', image: 'ubuntu18.04', test-type: 'unit', test-command: ${UNIT_TEST_CMD}, gpu: 'v100', driver: 'latest' }
               - { arch: 'arm64', python: '3.9', ctk: '11.8.0', image: 'ubuntu20.04', test-type: 'smoke', test-command: ${SMOKE_TEST_CMD}, gpu: 'a100', driver: 'latest' }
+              - { arch: 'arm64', python: '3.9', ctk: '12.0.1', image: 'ubuntu20.04', test-type: 'smoke', test-command: ${SMOKE_TEST_CMD}, gpu: 'a100', driver: 'latest' }
             nightly:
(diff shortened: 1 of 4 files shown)


.github/workflows/wheels-manylinux-test.yml
```diff
@@ -90,12 +90,18 @@ jobs:
           export MATRICES="
             pull-request:
               - { arch: 'amd64', python: '3.9', ctk: '11.8.0', image: 'ubuntu18.04', test-type: 'unit', test-command: ${UNIT_TEST_CMD}, gpu: 'v100', driver: 'latest' }
+              - { arch: 'amd64', python: '3.9', ctk: '12.0.1', image: 'ubuntu18.04', test-type: 'unit', test-command: ${UNIT_TEST_CMD}, gpu: 'v100', driver: 'latest' }
               - { arch: 'arm64', python: '3.9', ctk: '11.8.0', image: 'ubuntu20.04', test-type: 'smoke', test-command: ${SMOKE_TEST_CMD}, gpu: 'a100', driver: 'latest' }
+              - { arch: 'arm64', python: '3.9', ctk: '12.0.1', image: 'ubuntu20.04', test-type: 'smoke', test-command: ${SMOKE_TEST_CMD}, gpu: 'a100', driver: 'latest' }
```
ajschmidt8, line 95: I am going to open a follow-up PR to remove the `arm64` testing for wheels from `11.8` now that `12.0` is out.

We have very limited (8) `arm` GPU runners.

With this `arm` addition, a single PR will take up 4 GPUs:

- 1 arm GPU for conda C++ tests
- 1 arm GPU for conda Python tests
- 2 arm GPUs for wheel tests

This will cause really long queues for the arm GPU runners.

bdice: Can we do smoke tests instead of full tests?

ajschmidt8: That's what it's currently doing, but I still think it's impractical to have a single PR take up half of our ARM GPUs.

I think it's a fine trade-off to test only the latest CUDA version for ARM in PRs and rely on nightly testing for additional confidence.

We can always widen this matrix again in the future when we get more capacity.

PR #75, opened 2023-05-05
Support allowing wheels built for different CUDA versions in latter stages

divyegala:
This PR allows downstream repos to pass an un-evaluated string in a job that requires a wheel built from a previous job.

For example, currently commands are hard-coded as:
```
"RAPIDS_PY_WHEEL_NAME=pylibraft_cu11 ..."
```

Downstream repos will now be able to specify this command as:
```
"RAPIDS_PY_WHEEL_NAME=pylibraft_${{ '${PIP_CU_VERSION}' }} ..."
```

An upstream workflow now just needs to ensure `${PIP_CU_VERSION}` is set as `cu11/cu12/...` to download a wheel built from a previous job such that CUDA versions are correctly matched.

### DIFF
--- a/.github/workflows/wheels-manylinux-build.yml
+++ b/.github/workflows/wheels-manylinux-build.yml
@@ -203,6 +203,9 @@ jobs:
         pyenv global ${{ matrix.python }}
 
         # Run before-wheel command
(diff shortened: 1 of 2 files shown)


divyegala:
Here's passing wheel tests in a feature branch of RAFT that adds CUDA 12 support for wheels https://github.com/rapidsai/raft/actions/runs/4895920487/jobs/8742327631

ajschmidt8 (approved):

Issue #81, opened 2023-05-11
Move logic from citestwheel script into the test workflow


vyasr:
Currently wheel testing logic is contained in [a script embedded directly into the citestwheel image](https://github.com/rapidsai/cibuildwheel-imgs/blob/main/ciwheel/citestwheel.sh). The work here should proceed in two steps:
1. Move this logic from the image into the test workflows. 
2. Move the logic into scripts contained in each repo using the workflow. This step is only useful when we reach the stage of migrating the majority of workflow logic into scripts that can be handled by each repo using the workflow.

Would recommend tackling this after #79 since any attempt before then will duplicate code.

wence-:
1. is relatively straightforward, 2. runs into the potential trouble that now we're back to replicating boilerplate in all the repositories individually.

Possible proposal is that the `wheels-test` workflow sets up the environment and then the repo caller does something like:
```yaml
uses: wheels-test.yml
with:
   ....
run: |
    # set project-specific commands
    # call new gha-tools command that does equivalent of the old citestwheel script
```

vyasr:
I agree that there is going to be some boilerplate in every repo. I think that tradeoff is something that we are willing to accept to a degree in order to 1) make local repros possible, and 2) avoid any surprising or hidden behavior in the scripts used for local repros. Note that for conda builds the `ci/test_*` scripts in every repository are also somewhat duplicated. For instance, here are how [cuml](https://github.com/rapidsai/cuml/blob/branch-23.06/ci/build_cpp.sh) and [cugraph](https://github.com/rapidsai/cugraph/blob/branch-23.06/ci/build_cpp.sh) build the C++ conda packages. If you think that we can do even better without obfuscating anything I would be happy with that too, though. I'm happy to go over the `citestwheel.sh` script with you two if you'd like to see what logic exactly we anticipate getting duplicated; I think there's less truly boilerplate in there than there appears to be.

vyasr:
Closing, see https://github.com/rapidsai/shared-action-workflows/pull/90#issuecomment-1642525918 for explanation.

PR #90, opened 2023-05-16
Remove necessity for citestwheel.sh script in image

wence-:
This just inlines the contents and removes the no-longer-necessary definitions.

- Closes (partially) #81


.github/workflows/wheels-test.yml
```diff
@@ -152,11 +168,48 @@ jobs:
...
-        package-name: ${{ inputs.package-name }}
-
     - name: Run citestwheel
-      run: /citestwheel.sh
+      run: |
+        set -eoxu pipefail
+
+        ctk_tag="${{ matrix.ctk }}"
+        ctk_major=$(echo $ctk_tag | cut -d'.' -f1)
+        cuda_suffix="cu${ctk_major}"
```
vyasr, line 177: This can use [the new gha-tool that](https://github.com/rapidsai/gha-tools/pull/55) @divyegala added. [Here's how it's being used now](https://github.com/rapidsai/shared-action-workflows/pull/91/files).

.github/workflows/wheels-test.yml
```diff
@@ -152,11 +168,48 @@ jobs:
...
+        wheelname="${{ inputs.package-name }}_cu${ctk_major}"
+
+        export RAPIDS_PY_WHEEL_CUDA_SUFFIX=${cuda_suffix}
+        export RAPIDS_PY_WHEEL_NAME=${wheelname}
```
vyasr, line 181: Rather than exporting I think we can probably scope this narrowly to just the `rapids-download-wheels-from-s3` command.

.github/workflows/wheels-test.yml
```diff
@@ -152,11 +168,48 @@ jobs:
...
+                sh -c "${RAPIDS_BEFORE_TEST_COMMANDS_AMD64}"
+        elif [ "${arch}" == "aarch64" ]; then
+                sh -c "${RAPIDS_BEFORE_TEST_COMMANDS_ARM64}"
+        fi
+
+        # see: https://cibuildwheel.readthedocs.io/en/stable/options/#test-extras
+        extra_requires_suffix=''
+        if [ "${CIBW_TEST_EXTRAS}" != "" ]; then
+                extra_requires_suffix="[${CIBW_TEST_EXTRAS}]"
+        fi
```
vyasr, line 208: Also perhaps out of scope for this PR, but I'd like to for RAPIDS packages to expose the `[test]` extra as a matter of policy so that this just works without inputs.

vyasr (changes requested):
GH seems confused by the various upstream conflicts, may be easiest to re-review this after the upstream PR is merged. Generally looks good though, just a few comments.

vyasr:
This work has been superseded by the approach of moving test scripts directly into each repo. That was always the end goal, and it turns out that going straight to that is significantly less total work.

PR #94, opened 2023-05-23
Remove `sccache` config from wheel building workflow

divyegala:
Depends on https://github.com/rapidsai/cibuildwheel-imgs/pull/21

Closes https://github.com/rapidsai/shared-action-workflows/issues/82


vyasr (approved):
These changes look fine to me. I would like to see them run in practice on at least one repo more complex than rmm to verify that the build times still indicate sccache usage, so let's wait to merge until we have the new images built and can test.

PR #87, opened 2023-05-11
Add `GH_TOKEN` to conda build jobs

ajschmidt8:
This PR adds a `GH_TOKEN` to the conda build jobs to account for the changes in the PR below:

- https://github.com/rapidsai/gha-tools/pull/53

PR #97, opened 2023-05-23
Move wheel building logic to downstream repos

divyegala:
This PR tries to minimize custom logic in wheel building workflows by trying to mirror conda builds closer and moving wheel building logic to downstream repositories. The PR achieves this by adding a new workflow `wheels-build.yaml`

Depends on https://github.com/rapidsai/cibuildwheel-imgs/pull/29


vyasr (commented):
I'll hold off approving until everything is good on the downstream PRs, but this looks good to me.

.github/workflows/wheels-build.yml
```diff
@@ -0,0 +1,170 @@
+name: Build RAPIDS manylinux wheels
```
vyasr, line 1:
```suggestion
name: Build RAPIDS wheels
```

.github/workflows/wheels-build.yml
```diff
@@ -0,0 +1,170 @@
...
+          role-duration-seconds: 43200 # 12h
+      - name: checkout code repo
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ inputs.repo }}
+          ref: ${{ inputs.sha }}
+          fetch-depth: 0 # unshallow fetch for setuptools-scm
+          persist-credentials: false
+
+      - name: Standardize repository information
```
vyasr, line 138: Why can't we reuse `rapids-github-info` here? Is there some issue with whether or not the changes propagate to

.github/workflows/wheels-build.yml
```diff
@@ -0,0 +1,170 @@
...
+          ssh-key: ${{ secrets[inputs.extra-repo-deploy-key] }}
+          persist-credentials: false
+
+      - name: Build and repair the wheel
+        run: |
+          # Store internal pypi credentials before any step that may download wheels
+          printf 'machine pypi.k8s.rapids.ai\n\tlogin cibuildwheel\n\tpassword ${{ secrets.RAPIDSAI_PYPI_CI_PASSWORD }}\n' > ~/.netrc
+          ${{ inputs.build_script }}
+        env:
+          GH_TOKEN: ${{ github.token }}
```
vyasr, line 168: What is the token needed for? The private repo clone still happens in this workflow, is there another purpose?

divyegala: Not sure, `conda-cpp-build.yaml` also uses it so I coped it

.github/workflows/wheels-build.yml
divyegala: Yes we can, sorry I got confused because I picked it up from conda workflows

.github/workflows/wheels-build.yml
vyasr: Hmm looks like it was added in https://github.com/rapidsai/shared-action-workflows/pull/87 due to https://github.com/rapidsai/gha-tools/pull/53. Perhaps @ajschmidt8 could provide extra info, but I'm not sure that's needed for us because we never run `rapids-env-update`.

.github/workflows/wheels-build.yml
```diff
@@ -0,0 +1,166 @@
...
+          export MATRIX="
+          - { ctk: '11.8.0', arch: 'amd64', python: '3.9', linux_ver: 'centos7' }
+          - { ctk: '11.8.0', arch: 'amd64', python: '3.10', linux_ver: 'centos7' }
+          - { ctk: '11.8.0', arch: 'arm64', python: '3.9',  linux_ver: 'ubuntu20.04' }
+          - { ctk: '11.8.0', arch: 'arm64', python: '3.10', linux_ver: 'ubuntu20.04' }
+          - { ctk: '12.0.1', arch: 'amd64', python: '3.9',  linux_ver: 'centos7' }
+          - { ctk: '12.0.1', arch: 'amd64', python: '3.10',  linux_ver: 'centos7' }
+          - { ctk: '12.0.1', arch: 'arm64', python: '3.9', linux_ver: 'ubuntu20.04' }
+          - { ctk: '12.0.1', arch: 'arm64', python: '3.10', linux_ver: 'ubuntu20.04' }
+          "
```
ajschmidt8, line 98: can we make these matrix property names match the other workflows?

e.g. this one: https://github.com/rapidsai/shared-action-workflows/blob/614043874f8b8998e68a903db73bfb4253f78c3d/.github/workflows/conda-python-build.yaml#L55-L62

.github/workflows/wheels-build.yml
```diff
@@ -0,0 +1,166 @@
...
+
+          echo "MATRIX=$(
+            yq -n -o json 'env(MATRIX)' | \
+            jq -c '${{ inputs.matrix_filter }} | {include: .}' \
+          )" | tee --append "${GITHUB_OUTPUT}"
+
+  wheel-build:
+    name: cibuildwheel ${{ matrix.arch }} ${{ matrix.python }} ${{ matrix.ctk }}
```
ajschmidt8, line 106: We no longer use the `cibuildwheel` tool, right?

If so, it might make sense to remove it here to eliminate any ambiguity.

In fact, this `name` property is omitted entirely from our other workflows (see [here](https://github.com/rapidsai/shared-action-workflows/blob/614043874f8b8998e68a903db73bfb4253f78c3d/.github/workflows/conda-python-build.yaml#L68)). So we might be able to remove it here too.

.github/workflows/wheels-build.yml
ajschmidt8, line 1: Can we change this file extension to `.yaml` for consistency with most of the other workflows?

It looks like the `wheel` workflows deviated from this convention at some point.

.github/workflows/wheels-build.yml
```diff
@@ -0,0 +1,166 @@
...
+          - { ctk: '12.0.1', arch: 'arm64', python: '3.9', linux_ver: 'ubuntu20.04' }
+          - { ctk: '12.0.1', arch: 'arm64', python: '3.10', linux_ver: 'ubuntu20.04' }
+          "
+
+          echo "MATRIX=$(
+            yq -n -o json 'env(MATRIX)' | \
+            jq -c '${{ inputs.matrix_filter }} | {include: .}' \
+          )" | tee --append "${GITHUB_OUTPUT}"
+
+  wheel-build:
```
ajschmidt8, line 105:
```suggestion
  build:
```

In the context of a `wheels-build.yml` workflow, a job named `wheel-build` might be a bit redundant.

In other other workflows, we simply use `build` (see [here](https://github.com/rapidsai/shared-action-workflows/blob/614043874f8b8998e68a903db73bfb4253f78c3d/.github/workflows/conda-python-build.yaml#L68)) for succinctness.

.github/workflows/wheels-build.yml
ajschmidt8: I think it's fine to leave this in. The `gh` CLI tool can't be used without it. At some point someone may want to use the `gh` CLI tool in one of their scripts, at which point we'd have to add it back.

> Hmm looks like it was added in #87 due to rapidsai/gha-tools#53. Perhaps @ajschmidt8 could provide extra info, but I'm not sure that's needed for us because we never run rapids-env-update.

This is actually another opportunity for deduplication.

The `RAPIDS_DATE_STRING` variable and the `wheel-epoch-timestamp` job in this workflow are effectively doing the same thing: generating a unique per-workflow timestamp that can be used throughout the jobs in a given workflow.

I'm partial to using the `RAPIDS_DATE_STRING` method, which uses the `gh` CLI tool to get the timestamp of the start time of the currently running workflow. This method eliminates the need for an extra GHA job, which simplifies the workflow itself and the corresponding workflow graph that's generated in the GHAs web GUI.

I don't think we necessarily need to dedupe these implementations in this PR (though feel free if you think it's doable), but at the very least we should probably make a note of it.

@divyegala, @vyasr, what are your thoughts?

divyegala: @ajschmidt8 and I discussed this offline, and we think we can deduplicate and just create a small tool for `RAPIDS_DATE_STRING`. @vyasr this will also solve the problems we were discussing yesterday with producing `RAPIDS_EPOCH_TIMESTAMP` for local builds

vyasr (approved):

PR #99, opened 2023-05-24
Removes `auditwheel` `--plat` arg in favor of using `AUDITWHEEL_PLAT` env var in image

### LINKED ISSUE #80: Leverage AUDITWHEEL_PLAT in image to simplify auditwheel invocations
[This conditional logic](https://github.com/rapidsai/shared-action-workflows/blob/branch-23.06/.github/workflows/wheels-manylinux-build.yml#L231) can be replaced with a single call if the platform is specified in the container itself. [auditwheel respects the AUDITWHEEL_PLAT environment variable](https://github.com/pypa/auditwheel/pull/151). Unfortunately it's not documented anywhere else but [cibuildwheel relies on this](https://cibuildwheel.readthedocs.io/en/stable/options/#linux-image) so it should be fairly stable.

Depends on https://github.com/rapidsai/cibuildwheel-imgs/issues/16

rlratzel:
closes #80 
Merge after rapidsai/cibuildwheel-imgs#20

### DIFF
--- a/.github/workflows/wheels-manylinux-build.yml
+++ b/.github/workflows/wheels-manylinux-build.yml
@@ -237,11 +237,8 @@ jobs:
 
         # Repair the wheel
         cd dist
-        if [[ "${{ matrix.arch }}" == "amd64" ]]; then
-          python -m auditwheel repair -w . ${{ inputs.package-name }}* --plat manylinux_2_17_x86_64
-        else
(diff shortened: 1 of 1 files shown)


vyasr (approved):

vyasr:
The cibuildwheel-imgs PR is merged, the new images are uploaded, and CI on the [rmm testing PR](https://github.com/rapidsai/rmm/pull/1274) is now passing, so I think we're good to move forward here.

PR #103, opened 2023-05-31
Test Python 3.10 wheels / Ubuntu 20.04 in PR builds.

bdice:
This PR updates our CI matrix so that our "full" wheel tests cover a slightly wider range of platforms.

We test CUDA 11 and CUDA 12 on each PR, but we only cover Python 3.9 and Ubuntu 18.04. I propose that we test two more-orthogonal configurations with the full test suite (non-smoke tests):
```diff
  CUDA 11, Python 3.9, Ubuntu 18.04, amd64
- CUDA 12, Python 3.9, Ubuntu 18.04, amd64
+ CUDA 12, Python 3.10, Ubuntu 20.04, amd64
```

### DIFF
--- a/.github/workflows/wheels-manylinux-test.yml
+++ b/.github/workflows/wheels-manylinux-test.yml
@@ -90,7 +90,7 @@ jobs:
           export MATRICES="
             pull-request:
               - { arch: 'amd64', python: '3.9', ctk: '11.8.0', image: 'ubuntu18.04', test-type: 'unit', test-command: ${UNIT_TEST_CMD}, gpu: 'v100', driver: 'latest' }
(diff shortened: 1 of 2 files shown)


bdice:
@ajschmidt8 Are you okay with merging this directly or do we need to "migrate" every repo for this?

ajschmidt8:
I think this should be fine.

The CI image exists below and we build `3.10` and `3.9` wheels in PRs. So I don't anticipate any problems (famous last words though).

- https://hub.docker.com/r/rapidsai/citestwheel/tags?page=1&name=cuda12.0.1-ubuntu20.04-py3.10

ajschmidt8:
@bdice, I'll defer to you to merge and push this to the `cuda-120` branch.

ajschmidt8:
🤦‍♂️ thought I covered all the bases but I see now that I was looking at the wrong image.

PR #105, opened 2023-06-01
Use new wheel containers

divyegala:
https://github.com/rapidsai/cibuildwheel-imgs/pull/22 allowed for new wheel containers to be built with only one python version per container

### DIFF
--- a/.github/workflows/wheels-manylinux-build.yml
+++ b/.github/workflows/wheels-manylinux-build.yml
@@ -115,14 +115,14 @@ jobs:
           set -eo pipefail
 
           export MATRIX="
-          - { ctk: '11.8.0', arch: 'amd64', python: '3.9', image: 'rapidsai/manylinux_v2_2014:cuda-devel-11.8.0-centos7'}
-          - { ctk: '11.8.0', arch: 'amd64', python: '3.10', image: 'rapidsai/manylinux_v2_2014:cuda-devel-11.8.0-centos7' }
-          - { ctk: '11.8.0', arch: 'arm64', python: '3.9', image: 'rapidsai/manylinux_v2_2_31:cuda-devel-11.8.0-ubuntu20.04' }
-          - { ctk: '11.8.0', arch: 'arm64', python: '3.10', image: 'rapidsai/manylinux_v2_2_31:cuda-devel-11.8.0-ubuntu20.04' }
-          - { ctk: '12.0.1', arch: 'amd64', python: '3.9', image: 'rapidsai/manylinux_v2_2014:cuda-devel-12.0.1-centos7'}
-          - { ctk: '12.0.1', arch: 'amd64', python: '3.10', image: 'rapidsai/manylinux_v2_2014:cuda-devel-12.0.1-centos7' }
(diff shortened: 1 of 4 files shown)


.github/workflows/wheels-manylinux-build.yml
```diff
@@ -115,14 +115,14 @@ jobs:
...
           export MATRIX="
-          - { ctk: '11.8.0', arch: 'amd64', python: '3.9', image: 'rapidsai/manylinux_v2_2014:cuda-devel-11.8.0-centos7'}
-          - { ctk: '11.8.0', arch: 'amd64', python: '3.10', image: 'rapidsai/manylinux_v2_2014:cuda-devel-11.8.0-centos7' }
-          - { ctk: '11.8.0', arch: 'arm64', python: '3.9', image: 'rapidsai/manylinux_v2_2_31:cuda-devel-11.8.0-ubuntu20.04' }
-          - { ctk: '11.8.0', arch: 'arm64', python: '3.10', image: 'rapidsai/manylinux_v2_2_31:cuda-devel-11.8.0-ubuntu20.04' }
-          - { ctk: '12.0.1', arch: 'amd64', python: '3.9', image: 'rapidsai/manylinux_v2_2014:cuda-devel-12.0.1-centos7'}
-          - { ctk: '12.0.1', arch: 'amd64', python: '3.10', image: 'rapidsai/manylinux_v2_2014:cuda-devel-12.0.1-centos7' }
-          - { ctk: '12.0.1', arch: 'arm64', python: '3.9', image: 'rapidsai/manylinux_v2_2_31:cuda-devel-12.0.1-ubuntu20.04' }
-          - { ctk: '12.0.1', arch: 'arm64', python: '3.10', image: 'rapidsai/manylinux_v2_2_31:cuda-devel-12.0.1-ubuntu20.04' }
+          - { ctk: '11.8.0', arch: 'amd64', python: '3.9', linux_ver: 'centos7' }
```
vyasr, line 118: I noticed that these matrix variables are different from the corresponding conda jobs. Not something that we need to deal with now, but maybe worth unifying at some later point once we've merged the two container repos and moved more logic to repo-local scripts so that we can more fully align the wheel and conda workflows.

.github/workflows/wheels-manylinux-test.yml
```diff
@@ -118,7 +118,7 @@ jobs:
       matrix: ${{ fromJSON(needs.wheel-test-compute-matrix.outputs.MATRIX) }}
     runs-on: "linux-${{ matrix.arch }}-gpu-${{ matrix.gpu }}-${{ matrix.driver }}-1"
     container:
-      image: "rapidsai/citestwheel:cuda-devel-${{ matrix.ctk }}-${{ matrix.image }}"
+      image: "rapidsai/citestwheel:cuda${{ matrix.ctk }}-${{ matrix.image }}-py${{ matrix.python }}"
```
vyasr, line 121: This is the right change, just noting that this will potentially impact #90 CC @rlratzel

vyasr (approved):

PR #107, opened 2023-06-05
Remove `RAPIDS_DOCS_DIR` from `custom-job.yaml`

ajschmidt8:
In hindsight, this variable isn't needed at the workflow level since the only workflow step that uses it is the last one.

Therefore we can just set the variable's value in the script that's run in the last workflow step.


ajschmidt8:
xref https://github.com/rapidsai/shared-action-workflows/pull/93

PR #114, opened 2023-07-18
Add GH_TOKEN to env in custom-job.yaml

shwina:
It would be nice to be able to use the GH CLI in custom jobs, for which [this environment variable needs to be set](https://cli.github.com/manual/gh_auth_login)

### DIFF
--- a/.github/workflows/custom-job.yaml
+++ b/.github/workflows/custom-job.yaml
@@ -52,6 +52,7 @@ jobs:
     container:
       image: ${{ inputs.container_image }}
       env:
+        GH_TOKEN: ${{ github.token }}
         RAPIDS_BUILD_TYPE: ${{ inputs.build_type }}
         PARALLEL_LEVEL: ${{ env.PARALLEL_LEVEL }}
         NVIDIA_VISIBLE_DEVICES: ${{ env.NVIDIA_VISIBLE_DEVICES }}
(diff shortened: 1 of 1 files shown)


.github/workflows/custom-job.yaml
```diff
@@ -80,5 +81,6 @@ jobs:
           echo "RAPIDS_REF_NAME=${{ inputs.branch || github.ref_name }}" >> "${GITHUB_ENV}"
           echo "RAPIDS_REPOSITORY=${{ inputs.repo || github.repository }}" >> "${GITHUB_ENV}"
           echo "RAPIDS_SHA=$(git rev-parse HEAD)" >> "${GITHUB_ENV}"
+          echo "RAPIDS_PR_NUMBER=${{ fromJSON(steps.get-pr-info.outputs.pr-info).number }}" >> "${GITHUB_ENV}"
```
ajschmidt8, line 84:
```suggestion
```

This should be moved right beneath line 77 (in the `Add PR Info` step).

ajschmidt8 (changes requested):

ajschmidt8:
After some discussions offline with Ashwin, we will close this PR.

The reason we're closing is because we don't want to add `pull-requests: write` permissions.

Those permissions are only available to our PR workflows now since we copy pull request code to the source repository and therefore trigger our workflows on `push` events.

Once we move away from this method of testing PRs and use `pull_request` events, the maximum permissions we'll be able to supply the `pull-requests` key is `read` as described in the chart below

- https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

Therefore we don't want to make any changes to our permissions that further lock us into our current pull-request testing method since we do want to move away from it in the future.

PR #120, opened 2023-08-02
Remove old wheels workflows

vyasr:
All repos should have now been migrated to the new workflows.


bdice (approved):
LGTM. Let's merge this to avoid further confusion with old/new workflows.

bdice:
FYI: I verified that no repos were using the old workflows while filing #121.

PR #121, opened 2023-08-17
Remove old wheels workflows.

bdice:
It seems like the wheels "pure" and "manylinux" workflow files are no longer used. This PR removes them.

https://github.com/search?q=org%3Arapidsai+wheels-manylinux-build.yml&type=code
https://github.com/search?q=org%3Arapidsai+wheels-pure-build.yml&type=code

cc: @raydouglass @ajschmidt8 @vyasr @divyegala


bdice:
Closing in favor of #120. Can't believe I didn't check for an existing PR first. 🙄

PR #116, opened 2023-07-19
Move wheel testing logic to downstream repos

vyasr:
This PR is the testing companion to #97. It also introduces a standardized publish workflow that may be reused between pure and manylinux builds (see #89).

PR #132, opened 2023-09-22
Only match `.yaml` files in `update-version.sh`

raydouglass:
Since there are no `.yml` files in `.github/workflows`, the final `FILE` will be the literal `.github/workflows/*.yml` which is passed to `sed_runner` and fails.

Other repos do not try to match both extensions, but if we want to do that here, it would be more complex code using `find` and a read-while loop.

### DIFF
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -22,6 +22,6 @@ function sed_runner() {
     sed -i.bak ''"$1"'' $2 && rm -f ${2}.bak
 }
 
-for FILE in .github/workflows/*.{yaml,yml}; do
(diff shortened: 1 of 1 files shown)


ajschmidt8 (approved):
good catch.

I think this repo used to have both extensions, but we made them consistent when we were redoing some of the wheel workflows.

vyasr:
Yup see @ajschmidt8's request in https://github.com/rapidsai/shared-action-workflows/pull/97#discussion_r1268150812, and also in #116. The .yml files were removed in https://github.com/rapidsai/shared-action-workflows/pull/120.

PR #133, opened 2023-10-04
Revert "Update aws-actions/configure-aws-credentials action to v4"

bdice:
Reverts rapidsai/shared-action-workflows#130.

This update uses Node 20 which requires glibc 2.28 and thus does not work with CentOS 7 images.

### DIFF
--- a/.github/workflows/build-in-devcontainer.yaml
+++ b/.github/workflows/build-in-devcontainer.yaml
@@ -55,7 +55,7 @@ jobs:
             echo "HAS_DEVCONTAINER=false" >> "${GITHUB_ENV}";
           fi
       - if: ${{ env.HAS_DEVCONTAINER == 'true' }}
-        uses: aws-actions/configure-aws-credentials@v4
+        uses: aws-actions/configure-aws-credentials@v2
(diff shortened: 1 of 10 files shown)


.github/workflows/build-in-devcontainer.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
             echo "HAS_DEVCONTAINER=false" >> "${GITHUB_ENV}";
           fi
       - if: ${{ env.HAS_DEVCONTAINER == 'true' }}
-        uses: aws-actions/configure-aws-credentials@v4
+        uses: aws-actions/configure-aws-credentials@v2
```
bdice, line 58: Wondering if this should be v3.
```suggestion
        uses: aws-actions/configure-aws-credentials@v3
```

ajschmidt8 (approved):

ajschmidt8:
ugh. this will be fun.

GHAs are transitioning to Node20 and deprecating Node16

- https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

PR #136, opened 2023-10-05
Remove 11.4 tests

vyasr:
This removal was discussed on Slack, but would probably like approval from at least both @ajschmidt8 and @raydouglass before proceeding in case there is some other reason we need this job.

### DIFF
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -66,7 +66,6 @@ jobs:
           export MATRICES="
             pull-request:
               - { CUDA_VER: '11.2.2', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
-              - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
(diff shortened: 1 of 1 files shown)


bdice:
See also https://github.com/rapidsai/docs/pull/436#issuecomment-1749660780

We're planning to deprecate CUDA 11.2.

> Just to make sure we're on the same page - what will our new lower bound be (minimum support CUDA version) when we drop CUDA 11.2?

.github/workflows/conda-cpp-tests.yaml
```diff
@@ -62,37 +59,25 @@ jobs:
...
-              - { CUDA_VER: '11.5.2', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'earliest' }
-              - { CUDA_VER: '11.5.2', LINUX_VER: 'rockylinux8', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'earliest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
-            ext_nightly:
-              - { CUDA_VER: '11.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 't4', DRIVER: 'latest' }
```
bdice, line 86: Do we want to keep "extended nightly tests" but just update this version? The GPU is different here than any other jobs.

.github/workflows/conda-cpp-tests.yaml
```diff
@@ -62,37 +59,25 @@ jobs:
 
           export MATRICES="
             pull-request:
-              - { CUDA_VER: '11.2.2', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
```
raydouglass, line 65: Dropping this means no `centos7` or `earliest` driver coverage in PRs. Might want the `11.4` to use those?

.github/workflows/conda-cpp-tests.yaml
```diff
@@ -62,37 +59,25 @@ jobs:
...
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
             nightly:
-              - { CUDA_VER: '11.2.2', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
-              - { CUDA_VER: '11.2.2', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
-              - { CUDA_VER: '11.2.2', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
```
raydouglass, line 67: Related to the PR matrix comment:
```suggestion
              - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
```

.github/workflows/conda-cpp-tests.yaml
raydouglass: I don't think `extended_nightly_tests` is even used anywhere. Certainly not by nightlies in `rapidsai/workflows`.

ajschmidt8: It was supposed to be used for P100 and T4 testing, but we never enabled it.

We can remove it. If we want to enable additional GPU testing for select repositories in the future, we can re-add it.

vyasr: I did a search of the entire org before removing it and found no refs, which is why I thought it was safe. It sounds like Ray and AJ agree, so let's remove it for now. It's not hard to add back if needed.

.github/workflows/conda-cpp-tests.yaml
vyasr: I'm good with that.

vyasr: Done.

.github/workflows/conda-cpp-tests.yaml
vyasr: Done.

bdice (approved):

raydouglass (approved):

ajschmidt8 (approved):

PR #138, opened 2023-10-06
Use self-hosted runners for uploading packages

ajschmidt8:
GitHub-hosted runners have limited space which can cause issues with some jobs:

- https://github.com/rapidsai/cugraph/actions/runs/6428765880/attempts/1

This PR updates our shared workflows that are responsible for uploading wheel and conda packages to use self-hosted runners, which have significantly more space available.


ajschmidt8:
backport of #137

raydouglass (approved):

PR #140, opened 2023-10-10
Add ability run script to set a job_output that is re-exported

wence-:
If the run script sets job_output in the GITHUB_OUTPUTS environment variable this will now be exported as job_output to downstream dependants of this job.


wence-:
Needed for https://github.com/rapidsai/cudf-private/pull/61

targetted to 23.10 because that's the branch the xdf work is still based on

wence-:
Not generic enough to be worthwhile, so closing for now.

PR #144, opened 2023-10-18
Remove moved actions

AyodeAwe:
Removes the `get-pr-info` and `rapids-github-info` actions which have been implemented in the `shared-actions` repository. Also updates the code to reference that repository instead.


AyodeAwe:
refs should also be updated to use `main`. Pushing an update shortly

PR #145, opened 2023-10-18
Increase Docker's ulimit for maximum number of files

pentschev:
In some situations the default maximum number of files set in Docker may be insufficient, for example UCX-Py and UCXX have been hitting such limits frequently. We can increase those limits by adding specifying `container.options=--ulimit nofile=soft:hard`, and in this change we increase both soft and hard `ulimit`s to 1M.


ajschmidt8 (commented):
Glad this worked out!

I think this would be a good time to standardize the approach for passing container options across our three test jobs below:

- https://github.com/rapidsai/shared-workflows/blob/branch-23.12/.github/workflows/conda-cpp-tests.yaml
- https://github.com/rapidsai/shared-workflows/blob/branch-23.12/.github/workflows/conda-python-tests.yaml
- https://github.com/rapidsai/shared-workflows/blob/branch-23.12/.github/workflows/wheels-test.yaml

Can we follow the convention outlined in `wheels-test.yaml` shown below, where the test workflow accepts an input that is passed to the container options?

- https://github.com/rapidsai/shared-workflows/blob/cbd604b1668e11a8f1b3dc39d0d5d823f3c5cf52/.github/workflows/wheels-test.yaml#L27-L31


I think a more suitable/succint name for that input would be `container-options`.

Changing this name in the `wheels-test.yaml` workflow would technically be a breaking change, but from the query below, it seems that the option is only used in the `ucx-py` repository, which means it should be pretty easy to coordinate the update across repos.

- https://github.com/search?q=org%3Arapidsai%20test-docker-options&type=code

What do you think of these proposals?

pentschev:
@ajschmidt8 I've now made the changes you proposed along with the PRs to apply the new limits as we need:

- https://github.com/rapidsai/ucx-py/pull/1002
- https://github.com/rapidsai/ucxx/pull/106

Could you take a look when you have a chance?

ajschmidt8 (approved):

PR #148, opened 2023-10-25
Update actions for branch cuda-120-arm

AyodeAwe:
Removes the `get-pr-info` and `rapids-github-info` actions which have been implemented in the `shared-actions` repository. Also updates the code to reference that repository instead.

Repeats the changes made [here](https://github.com/rapidsai/shared-workflows/pull/144) for branch `cuda-120-arm`. This is necessary because the `cuda-120-arm` branch is being referenced in places like [here](https://github.com/rapidsai/wholegraph/blob/branch-23.12/.github/workflows/build.yaml).

See this slack [thread](https://nvidia.slack.com/archives/C03KUM1L2G4/p1698190050388989) for more.


AyodeAwe:
Will close this as it has been determined that we should switch the dependent library's (`wholegraph`) reference to this branch `cuda-120-arm` to `branch-23.12`.

PR #150, opened 2023-10-27
Add option for additional build jobs

vyasr:
This PR adds an option for including additional build jobs in the matrix. These are built on new the rockylinux8 ci-wheel images added in https://github.com/rapidsai/ci-imgs/pull/87.

I'm not sure what we want to do for tests; the current test matrix means that some of the 2.28 wheels could easily be tested by just changing the download logic since they have new enough OSes to prefer the new wheels, but now it'll be an even smaller subset of the total builds. Also the nightly run won't test everything anymore. If we want to be thorough we'll have to double the number of jobs in nightlies to run on both old and new glibc citestwheel images.

### DIFF
--- a/.github/workflows/wheels-test.yaml
+++ b/.github/workflows/wheels-test.yaml
@@ -67,7 +67,7 @@ jobs:
           export MATRICES="
             pull-request:
               - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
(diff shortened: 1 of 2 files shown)


ajschmidt8 (approved):

ajschmidt8 (approved):

.github/workflows/wheels-test.yaml
```diff
@@ -67,7 +67,7 @@ jobs:
           export MATRICES="
             pull-request:
               - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
```
bdice, line 70: @vyasr Did we want to make a similar change for the nightly matrix, maybe on line 78?

PR #151, opened 2023-10-30
Switch default arm builds to rockylinux8

vyasr:
Prior to #150, arm builds were only done using ubuntu20.04, which uses glibc 2.31. This was based on the available set of docker/cuda images at the time we started building wheels. Since then, rockylinux8 images have been added, and to support #150 we added corresponding ci-wheel images. #150 actually doubled the number of wheels being built without good reason. There is no need to build 2.28 and 2.31 wheels, we can simply start building 2.28 wheels on arm as the new default and drop the 2.31 wheels altogether.


vyasr:
If desired, we could instead use the rockylinux builds as the new default on arm to slightly expand our support matrix to a minimum glibc of 2.28.

ajschmidt8 (approved):

PR #152, opened 2023-10-31
Conda cpp post build checks

robertmaynard:
Introduce the concept of post C++ build checks. Includes the first optional check which scans a the produced C++ conda libraries for any weak CUDA kernels.

### DIFF
--- a/.github/workflows/conda-cpp-post-build-checks.yaml
+++ b/.github/workflows/conda-cpp-post-build-checks.yaml
@@ -67,8 +67,9 @@ jobs:
           echo "RAPIDS_NIGHTLY_DATE=${{ inputs.date }}" >> "${GITHUB_ENV}"
       - name: Download conda C++ build artifacts
         run: |
-          rapids-download-conda-from-s3 cpp
-          rapids-extract-conda-files cpp
+          CPP_DIR=$(rapids-download-conda-from-s3 cpp)
+          EXTRACTED_DIR=$(rapids-extract-conda-files "${CPP_DIR}")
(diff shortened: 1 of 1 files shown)


robertmaynard:
Verification runs of this workflow on raft:
Job passes as expected when given the correct symbol exclusion list: https://github.com/rapidsai/raft/actions/runs/6697280824/job/18198270374

Job fails as expected when the default symbol exclusion list is used:
https://github.com/rapidsai/raft/actions/runs/6699928979/job/18205855860?pr=1935

.github/workflows/conda-cpp-post-build-checks.yaml
```diff
@@ -0,0 +1,81 @@
...
+  repository-projects: none
+  security-events: none
+  statuses: none
+
+jobs:
+  check-symbols:
+    if: ${{ inputs.enable_check_symbols }}
+    runs-on: linux-amd64-cpu4
+    container:
+      image: rapidsai/ci-wheel:cuda12.0.1-centos7-py3.10
```
ajschmidt8, line 47:
```suggestion
      image: rapidsai/ci-wheel:latest
```

Does the version of this container matter?

If not, can we switch it to `latest`?

The versions used for our `latest` tags are specified in the `rapidsai/ci-imgs` repository here: https://github.com/rapidsai/ci-imgs/blob/main/latest.yaml

ajschmidt8 (commented):
Just one question

.github/workflows/conda-cpp-post-build-checks.yaml
robertmaynard: The version doesn't matter. So moving to use `latest`

ajschmidt8 (approved):

PR #154, opened 2023-11-13
Upload wheels to public anaconda index

AyodeAwe:
PR updates the wheels-publish step to upload wheels to anaconda instead of the existing internal PyPI.

Cross-references:
- https://github.com/rapidsai/ci-imgs/pull/92
- https://github.com/rapidsai/gha-tools/pull/86

### DIFF
--- a/.github/workflows/build-in-devcontainer.yaml
+++ b/.github/workflows/build-in-devcontainer.yaml
@@ -93,7 +93,7 @@ jobs:
             mkdir -m 0775 -p ~/.config/pip/;
             cat <<EOF >> ~/.config/pip/pip.conf
             [global]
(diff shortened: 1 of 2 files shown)


vyasr (commented):
Will we have to propagate the token to all repos that upload wheels? Have we done that yet?

vyasr:
While we're at it, we should also update https://github.com/rapidsai/shared-workflows/blob/branch-23.12/.github/workflows/build-in-devcontainer.yaml#L96 so that devcontainer builds look in the new index.

AyodeAwe:
> Will we have to propagate the token to all repos that upload wheels? Have we done that yet?

Good catch. Looking into that now

AyodeAwe:
> Will we have to propagate the token to all repos that upload wheels? Have we done that yet?

Yep. They are propagated to all repos that upload wheels.

PR #158, opened 2023-11-21
Switch to 20.04 for one nightly job to better test glibc compatibility

vyasr:
This change was suggested by @bdice [here](https://github.com/rapidsai/shared-workflows/pull/150#discussion_r1376555395) and ensures that at least one nightly job validates wheels in cases where we build for multiple manylinux versions. Currently we only do that for PRs (as of #150).


bdice (approved):
LGTM. To clarify: this expands the matrix of glibc versions tested (and thus the manylinux tags pulled) _for amd64 architectures_. The arm64 jobs are already using Ubuntu 20.04, and older versions aren't supported iirc.

ajschmidt8 (approved):

PR #160, opened 2023-11-29
Remove k8s creds

vyasr:
These credentials are no longer needed since we're on the public index.


ajschmidt8 (approved):
:tada: :tada: :tada:

divyegala (approved):

PR #166, opened 2023-12-20
add CUDA 12.2 conda and wheel jobs

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/7.
Blocked by https://github.com/rapidsai/ci-imgs/pull/101.

Proposes adding CUDA 12.2.2 jobs for conda and wheel building and testing.

### Why these specific matrix entries?

*(updated Jan 10)*

For *building*:

* dropping CUDA 12.0 and replacing it with 12.2 because:
    - artifact names are of the form `cuda{major}`, so otherwise-identical builds with different minor versions of CUDA would overwrite each others' wheels (https://github.com/rapidsai/shared-workflows/pull/166#discussion_r1439749041)
    - CUDA Enhanced Compatibility (CEC) mean you should be able to build against CUDA 12.2 and run with CUDA 12.0 (https://github.com/rapidsai/shared-workflows/pull/166#discussion_r1439743881)

For *testing*

* dropping a few 12.0 `arm64` because our `arm64` capacity is more limited than `x86_64` (https://github.com/rapidsai/shared-workflows/pull/166#discussion_r1439751864)
* preserving most 12.0 tests to ensure relying on CEC is giving us what we expect it to (https://github.com/rapidsai/shared-workflows/pull/166#discussion_r1439743881)
* using `(ubuntu20.04, ubuntu22.04)` (leaving `(ubuntu18.04, ubuntu20.04)`)) for 12.2 tests, as CUDA 12.2 doesn't support Ubuntu 18.04 (https://github.com/rapidsai/shared-workflows/pull/166#discussion_r1439758006)

### Why not use `continue-on-error`?

That idea came from a desire to avoid needing to do this "stage a PR in every RAPIDS repo" approach to testing this.

The fact that artifacts like wheels have names that only contain the CUDA *major* version means that having builds against both CUDA 12.0 and CUDA 12.2 would require hacking in some changes to how artifacts are named and then ensuring that testing knows about those changes (https://github.com/rapidsai/shared-workflows/pull/166#discussion_r1439749041). That's made more complicated by the fact that many projects have repo-local scripts that do artifact uploading and downloading (https://github.com/rapidsai/shared-workflows/pull/166#discussion_r1442182995).

Given that, I'm proposing just doing this with the "stage a PR in every RAPIDS repo" approach. And given that, `continue-on-error` doesn't give us any benefit but WOULD introduce an elevated risk of missing CI failures.

### Why not a branch from a fork?

So I can test workflows in other repos like `rmm` and `cudf` using the workflows from this branch.

### Rollout plan?

1. [x] get images from https://github.com/rapidsai/ci-imgs/pull/101 built and published
2. [x] test this branch on all RAPIDS repos one at a time, in the following order (based on the graph I see at https://github.com/rapidsai/workflows/actions/runs/7405716125), and asking for help from those projects' maintainers if things fail
    * moved list to https://github.com/rapidsai/build-planning/issues/7#issuecomment-1887577440
4. [ ] merge this branch into `branch-24.04`
5. [ ] another round of PRs to all those projects and switch them back from this branch to `branch-24.04` for these workflows
6. [ ] delete this branch

### DIFF
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -73,7 +73,6 @@ jobs:
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'arm64', PY_VER: '3.9', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
-              - { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -109,6 +109,10 @@ jobs:
           aws-region: ${{ vars.AWS_REGION }}
           role-duration-seconds: 43200 # 12h
       - uses: actions/checkout@v3
+        env:
+          GIT_TRACE_PACKET: 1
+          GIT_TRACE: 1
+          GIT_CURL_VERBOSE: 1
         with:
           repository: ${{ inputs.repo }}
           ref: ${{ inputs.sha }}
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -65,8 +65,8 @@ jobs:
             pull-request:
               - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
-              - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
-              - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
+              - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
+              - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
             nightly:
               - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -66,17 +66,18 @@ jobs:
               - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
-              - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
+              - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
             nightly:
               - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'arm64', PY_VER: '3.9', GPU: 'a100', DRIVER: 'latest' }
-              - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
+              - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
+              - { CUDA_VER: '12.2.2', LINUX_VER: 'rockylinux8', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
           "
 
           TEST_MATRIX=$(yq -n 'env(MATRICES) | .[strenv(BUILD_TYPE)]')
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -68,18 +68,19 @@ jobs:
             pull-request:
               - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
-              - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
+              - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
             nightly:
               - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'arm64', PY_VER: '3.9', GPU: 'a100', DRIVER: 'latest' }
-              - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
+              - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
+              - { CUDA_VER: '12.2.2', LINUX_VER: 'rockylinux8', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
           "
 
           TEST_MATRIX=$(yq -n 'env(MATRICES) | .[strenv(BUILD_TYPE)]')
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -76,7 +76,6 @@ jobs:
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'arm64', PY_VER: '3.9', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
-              - { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
               - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -113,6 +113,10 @@ jobs:
           aws-region: ${{ vars.AWS_REGION }}
           role-duration-seconds: 43200 # 12h
       - uses: actions/checkout@v3
+        env:
+          GIT_TRACE_PACKET: 1
+          GIT_TRACE: 1
+          GIT_CURL_VERBOSE: 1
         with:
           repository: ${{ inputs.repo }}
           ref: ${{ inputs.sha }}
--- a/.github/workflows/wheels-build.yaml
+++ b/.github/workflows/wheels-build.yaml
@@ -130,6 +130,10 @@ jobs:
           role-duration-seconds: 43200 # 12h
       - name: checkout code repo
         uses: actions/checkout@v3
+        env:
+          GIT_TRACE_PACKET: 1
+          GIT_TRACE: 1
+          GIT_CURL_VERBOSE: 1
         with:
           repository: ${{ inputs.repo }}
           ref: ${{ inputs.sha }}
@@ -154,6 +158,10 @@ jobs:
       - name: checkout extra repos
         uses: actions/checkout@v3
         if: ${{ inputs.extra-repo != '' }}
+        env:
+          GIT_TRACE_PACKET: 1
+          GIT_TRACE: 1
+          GIT_CURL_VERBOSE: 1
         with:
           repository: ${{ inputs.extra-repo }}
           ref: ${{ inputs.extra-repo-sha }}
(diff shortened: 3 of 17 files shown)


.github/workflows/wheels-build.yaml
```diff
@@ -91,13 +91,19 @@ jobs:
           - { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.10',  LINUX_VER: 'centos7' }
           - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.9', LINUX_VER: 'rockylinux8' }
           - { CUDA_VER: '12.0.1', ARCH: 'arm64', PY_VER: '3.10', LINUX_VER: 'rockylinux8' }
+          - { CUDA_VER: '12.2.2', ARCH: 'amd64', PY_VER: '3.9',  LINUX_VER: 'centos7' }
+          - { CUDA_VER: '12.2.2', ARCH: 'amd64', PY_VER: '3.10',  LINUX_VER: 'centos7' }
+          - { CUDA_VER: '12.2.2', ARCH: 'arm64', PY_VER: '3.9', LINUX_VER: 'rockylinux8' }
+          - { CUDA_VER: '12.2.2', ARCH: 'arm64', PY_VER: '3.10', LINUX_VER: 'rockylinux8' }
```
bdice, line 97: We only need to _build_ with one CUDA version from each major version. This is because of CUDA Enhanced Compatibility (CEC), which ensures that builds for CUDA 12.x are backwards compatible with CUDA 12.y at runtime (`y < x`). https://docs.nvidia.com/deploy/cuda-compatibility/

(Also, the wheels are labeled by CUDA major version like `-cu12`, not major.minor, so we wouldn't even be able to distinguish the 12.0 and 12.2 builds. This is intentional, because the guarantees we get from CEC mean that we don't need CUDA minor versions in the wheel names.)

We can drop builds with 12.0.1 but I would vote to retain (some? all?) tests for 12.0.1. In the past we've expressed the intention to test according to this policy (previously discussed with @raydouglass @ajschmidt8, but this policy may not be documented?):

- latest CUDA from the previous major version (11.8.0)
- earliest CUDA from the current major version (12.0.1)
- latest CUDA from the current major version (now 12.2.2) (* see below!)

Until this PR added 12.2, the last two bullets were the same (we built with 12.0 and tested with 12.0 so testing CEC was irrelevant since the build/test matrices matched). Adding 12.2 means that we want to add a test config for 12.2 while retaining 12.0 for CEC, to align with this policy.

The asterisk on the last one is that RAPIDS typically lags CUDA minor releases by a bit. For example: we could test with 12.3 in theory, but we don't have GPU CI runners with new enough drivers yet. We are sticking to 12.2 for now because it's the latest CUDA minor version that has a corresponding LTS driver branch (R535). https://docs.nvidia.com/datacenter/tesla/drivers/index.html#cuda-drivers

.github/workflows/wheels-build.yaml
```diff
@@ -112,6 +118,7 @@ jobs:
   build:
     name:  ${{ matrix.CUDA_VER }}, ${{ matrix.PY_VER }}, ${{ matrix.ARCH }}, ${{ matrix.LINUX_VER }} 
     needs: [compute-matrix]
+    continue-on-error: ${{ matrix.CUDA_VER == '12.2.2' }}
```
bdice, line 121: > ### Why not a branch from a fork?
> So I can test workflows in other repos like rmm and cudf using the workflows from this branch.

FYI -- if we're committed to trying the "continue on error" approach, it won't matter if this PR comes from a fork. We would merge it to the upstream `rapidsai/shared-workflows`, and just let the jobs fail. The only time it matters if it comes from a fork is for using branch name updates in each repo's actions that we revert after completing the migration.

bdice: However, it gets a little weird for this case of updating CUDA 12.2 if we were to merge this PR with `continue-on-error`, because we would like to test the artifacts from both 12.0 (the status quo) and 12.2 (the new version) but the artifacts will have the same name `-cu12`. I'm not sure how we would distinguish this while making the updates non-breaking. For example, dropping CUDA 12.0 builds at the same time would be breaking if the new 12.2 build jobs don't pass CI, and it wouldn't even make sense to do `continue-on-error`. We'd have to add builds for 12.2 that _don't_ upload artifacts, make sure they pass, and then disable 12.0 at the same time we make the 12.2 builds upload their artifacts. Unless there's a way to change the names of the 12.2 artifacts and test those explicitly...? This needs some consideration.

.github/workflows/wheels-test.yaml
```diff
@@ -69,6 +69,8 @@ jobs:
               - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.9', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
+              - { ARCH: 'arm64', PY_VER: '3.9', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
```
bdice, line 73: Since the other arm64 job for CUDA 12.0.1 is testing Python 3.9, it would be ideal to test Python 3.10 here (spread the coverage across the matrix).
```suggestion
              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
```

However, it's important to note that our arm64 resources are limited and we have an exception to the policy I mentioned above. We don't have any PR tests for CUDA 11 on ARM, so that implies that we have a goal of running only one ARM job for wheels per pull request. If so, we could only test CUDA 12.2 and skip testing 12.0 (so the number of ARM test jobs stays constant). And if it's true that we only want one ARM CI job for wheels in pull requests, then I would prefer to only test the latest supported Python version (currently 3.10).

.github/workflows/wheels-test.yaml
```diff
@@ -78,6 +80,10 @@ jobs:
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.9', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
```
bdice, line 83: CUDA 12.2.2 doesn't list Ubuntu 18.04 in the support matrix.

Compare:
https://docs.nvidia.com/cuda/archive/12.0.1/cuda-installation-guide-linux/index.html#os-support-policy
https://docs.nvidia.com/cuda/archive/12.2.2/cuda-installation-guide-linux/index.html#os-support-policy

This agrees with the Docker image support. (In the past there have been some discrepancies.) There is no `nvidia/cuda` image for CUDA 12.2 on Ubuntu 18.04:
https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=12.0.1-runtime-ubuntu
https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=12.2.2-runtime-ubuntu

I would drop this line.
```suggestion
```

We also probably need to evaluate whether we should be testing Ubuntu 22.04 somewhere in this matrix.

.github/workflows/wheels-build.yaml
bdice: Just checking myself on this -- I probably need to refresh myself on how we determine which wheel artifacts to publish. We can build for 12.2 and not publish, and we can test the 12.0 builds on 12.2 drivers. I just don't know how that works in practice because I am not 100% sure whether this step would run into conflicting names from the 12.0 wheels and 12.2 wheels: https://github.com/rapidsai/shared-workflows/blob/4384050a538188eb6f4d19cb3865cdecb86b3c46/.github/workflows/wheels-build.yaml#L181

jameslamb: >  if we're committed to trying the "continue on error" approach, it won't matter if this PR comes from a fork

This makes sense to me in general.

I wanted to do this from a branch here *this time* because I'm also testing my use of the `continue-on-error` mechanism itself. I was nervous about misunderstanding that mechanism and merging something that breaks CI in a merge-blocking way for a bunch of projects all at once.

> This needs some consideration.

Definitely, let me think on that! 🤔 

I forgot that the wheels will have package names (and therefore filenames) like `{project}-cu{major}`.

.github/workflows/wheels-test.yaml
jameslamb: Ok yeah that makes sense to me. I ran into that lack of Ubuntu 18.04 images over in https://github.com/rapidsai/ci-imgs/pull/101#discussion_r1432975445 (which is upstream of this PR), but I wasn't sure whether that was intentional. Those OS Support Policy docs answer that for me, thank you!

.github/workflows/wheels-build.yaml
jameslamb: > I am not 100% sure whether this step would run into conflicting names from the 12.0 wheels and 12.2 wheels

Your intuition was right... it would run into conflicts.

That statement you linked to trims out the first `.` and everything after it.

```shell
RAPIDS_CUDA_VERSION="12.2.2"
echo "${RAPIDS_CUDA_VERSION%%.*}"
# 12
```

The CUDA *major* version is baked into artifact names and not used to disambiguate objects in S3 (see https://github.com/rapidsai/ops/issues/2982#issuecomment-1861157155 for reference).

e.g. see https://downloads.rapids.ai/nightly/cudf/2024-01-04/4c01e95/

Screenshot 2024-01-04 at 11 49 51 AM

So as this is all currently set up, having `build-*` steps with different CUDA major versions would result in artifacts conflicting.

I'm thinking about this 🤔

jameslamb: Individual projects are also uploading and downloading artifacts from code inside repo-local scripts.

For example, from `rmm`:

```yaml
  wheel-build:
    needs: checks
    secrets: inherit
    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-24.02
    with:
      build_type: pull-request
      script: ci/build_wheel.sh
```

([pr.yaml](https://github.com/rapidsai/rmm/blob/e2dc7279d9a6bed2860697bcb5155a0ee6e57efe/.github/workflows/pr.yaml#L65-L71))

```shell
package_name="rmm"

# ...

RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"

# ...

RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 final_dist
```

([build_wheel.sh](https://github.com/rapidsai/rmm/blob/e2dc7279d9a6bed2860697bcb5155a0ee6e57efe/ci/build_wheel.sh#L39))

.github/workflows/wheels-test.yaml
bdice: @jameslamb What are your thoughts here? Keep this to one ARM job? If so, should we test only Python 3.10?

.github/workflows/wheels-test.yaml
```diff
@@ -68,16 +68,20 @@ jobs:
...
               - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
-              - { ARCH: 'arm64', PY_VER: '3.9', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', gpu: 'v100', driver: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
+              - { ARCH: 'arm64', PY_VER: '3.9', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
+              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', gpu: 'a100', driver: 'latest' }
```
bdice, line 84: If we wanted to keep this to two ARM jobs, we could reduce this to:
- ARM, Python 3.9, CUDA 12.0, Ubuntu 20.04
- ARM, Python 3.10, CUDA 12.2, Ubuntu 22.04
```suggestion
              - { ARCH: 'arm64', PY_VER: '3.9', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
              - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', gpu: 'v100', driver: 'latest' }
              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', gpu: 'a100', driver: 'latest' }
```

bdice (approved):
LGTM, up to the questions about ARM testing in PRs.

.github/workflows/wheels-test.yaml
jameslamb: blegh I made a mistake here. I agree with you it should be:

* only 1 `arm` job on PRs
* that job should use CUDA 12.2 + Python 3.10

I meant to do that, actually... that job being Python 3.9 here was a mistake. Just pushed https://github.com/rapidsai/shared-workflows/pull/166/commits/b3911076b7c2cf88729389495fd1831a20099eaf fixing that.

.github/workflows/wheels-test.yaml
jameslamb: Yeah I'm ok with that. I wasn't sure where to draw the line between expanded coverage vs. minimizing runner capacity.

@vyasr 's comment over in https://github.com/rapidsai/build-planning/issues/5 gives me more confidence that we're not giving up much by keeping the number of `arm64` job constant in this PR.

> *We very rarely encounter cases where our CI tests uncover a bug that is only present in one Python version.*

Thanks!

.github/workflows/wheels-test.yaml
jameslamb: resolving this thread... I dropped this failing ubuntu18.04 build and added a CUDA 12.2 + ubuntu22.04 one (see PR description).

.github/workflows/wheels-build.yaml
jameslamb: We decided to address this by abandoning the `continue-on-error` approach and instead doing the "open PRs on all the RAPIDS repos" approach. That seemed simpler than messing around with artifact-naming.

See the PR description for more details.

.github/workflows/wheels-build.yaml
ajschmidt8: I would vote to omit `12.0.1` from the PR test matrix for now so that we can keep the number of combinations to three.

Once we get on the new GPU cluster, we'll have more capacity and can expand the matrix. Until then, we can just increase the nightly matrix.

I worry that adding too many combinations to the PR tests will increase our queue times too much given the number of repositories that now use our shared workflows.

jameslamb: > we can keep the number of combinations to three

That's fine with me (you know about our available runner capacity a lot better than me), but how about we accomplish that by instead having 1 CUDA 12.0 job and 1 CUDA 12.2 job for the PR test matrix?

With a mix of Python versions and architecture, like this:

```text
- { CUDA_VER: '12.0.1', ARCH: 'amd64', PY_VER: '3.9',  LINUX_VER: 'rockylinux8' }
- { CUDA_VER: '12.2.2', ARCH: 'arm64', PY_VER: '3.10',  LINUX_VER: 'rockylinux8' }
```

That'd keep the absolute number of jobs the same, but give us some coverage of both 12.0 and 12.2 on PRs, so that issues with relying on CUDA Enhanced Compatibility could be caught ealier.

jameslamb: > I think we've concluded in the past with Vyas that it's better to test across CUDA versions than Python versions in PRs

Perfect, yeah I was thinking along those lines because of what he wrote at https://github.com/rapidsai/build-planning/issues/5.

> So then we'll drop one of the 11.8.0 jobs?

No, there's already only 1 11.8.0 job being run on PRs ([code link](https://github.com/rapidsai/shared-workflows/blob/91799a905608f4c57d7fe65b92ce9a261249e635/.github/workflows/wheels-test.yaml#L68-L71))

```text
- { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
- { ARCH: 'arm64', PY_VER: '3.9', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
```

And 3 total jobs.

I'm proposing preserving that 11.8.0 job, removing one of the 12.0.1 jobs, and adding one 12.2.0 job, like this:

```text
- { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
- { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
```

Then CUDA 11.8.0, 12.0.1, and 12.2.2 would be tested on every PR. Without any change in the absolute number of `wheels-test` jobs per commit on PRs.

jameslamb: I just pushed that change in https://github.com/rapidsai/shared-workflows/pull/166/commits/b4f97a69e2db7839941406647a1d6524c8a568e5.

So as of the current state of this branch, the PR test matrix:

* has 3 total jobs (just like on `branch-24.02`)
* has 1 job for each of CUDA 11.8, 12.0, 12.2
* spreads architecture, Python version, and Ubuntu version over those

.github/workflows/wheels-test.yaml
```diff
@@ -67,8 +67,8 @@ jobs:
           export MATRICES="
             pull-request:
               - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
-              - { ARCH: 'arm64', PY_VER: '3.9', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
+              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
```
bdice, line 71: - Let's favor Python 3.10 in 2/3 of the jobs rather than 3.9
- Let's test amd64 arch with "newest everything" (rather than arm64) as it's the most common combination of features being used in the real world
  - Therefore, leave arm64 with CUDA 12.0 for version coverage

```suggestion
              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
```

.github/workflows/wheels-build.yaml
bdice: Great. I like this. I proposed a small tweak here: https://github.com/rapidsai/shared-workflows/pull/166/files#r1445403457

.github/workflows/wheels-test.yaml
jameslamb: works for me!

ajschmidt8 (approved):

.github/workflows/conda-cpp-tests.yaml
```diff
@@ -65,18 +65,19 @@ jobs:
...
               - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
-              - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
+              - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
             nightly:
               - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'arm64', PY_VER: '3.9', GPU: 'a100', DRIVER: 'latest' }
-              - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
```
jameslamb, line 74: This exact combination is already being tested on every PR, so thought it was safe to remove. With this removal, the net effect for `cpp-tests` workflow is:

* PRs: no change
* nightly: +1 job per run

.github/workflows/conda-python-tests.yaml
```diff
@@ -68,18 +68,19 @@ jobs:
...
               - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
               - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
-              - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
+              - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
             nightly:
               - { CUDA_VER: '11.4.3', LINUX_VER: 'centos7', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'amd64', PY_VER: '3.9', GPU: 'v100', DRIVER: 'latest' }
               - { CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', ARCH: 'arm64', PY_VER: '3.9', GPU: 'a100', DRIVER: 'latest' }
-              - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
```
jameslamb, line 77: Similar to https://github.com/rapidsai/shared-workflows/pull/166/files#r1447666026 .... removing this to limit impact on CI capacity, and since this combination is already tested on PRs.

jameslamb:
@ajschmidt8 FYI ... yesterday, @bdice decided to try to expand this to doing `conda` packages and wheels at the same time. For the following reasons:

1. reduces the total testing effort (e.g. no need to do another round of PRs to every RAPIDS repo to test `conda` separately from wheels)
2. because CUDA 12.2 packages are all now available on `conda-forge` (https://github.com/conda-forge/cuda-feedstock/issues/13)
3. to make it easier to roll back all of this as one atomic unit if it doesn't get done before code freeze for `24.02`

I just pushed commits here proposing matrix changes for `conda-*-build` and `conda-*-tests` workflows. Once we agree on those changes, I'll open the remaining PRs to test in all RAPIDS repos so we can see what types of problems come up.

ajschmidt8:
Looks good to me!

.github/workflows/conda-cpp-tests.yaml
bdice: I wonder if we could shed one of these nightly jobs to make it net-zero for ARM runners. ARM runner capacity is the tightest constraint for us.
```
- { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
- { CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', ARCH: 'arm64', PY_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
```

.github/workflows/conda-python-tests.yaml
bdice: Same question as above about dropping one 11.8 ARM job.

.github/workflows/conda-cpp-tests.yaml
jameslamb: Sure! I just pushed https://github.com/rapidsai/shared-workflows/pull/166/commits/8fcd98f7e0b583df6e0dcef4df5ee921025f2b2a removing one of them from each `conda-*-test` workflow.

Chose entries that'd ensure we still have coverage of operating systems (e.g. 1 `arm64-ubuntu22.04` and 1 `arm64-rockylinux8`).

.github/workflows/conda-python-tests.yaml
jameslamb: done: https://github.com/rapidsai/shared-workflows/pull/166#discussion_r1447745674

vyasr:
Everything looks good so far. Let me know if you run into any snags in the testing phase.

The discussions around wheel naming and how it prevents us from building both 12.0 and 12.2 wheels makes sense. Reworking `rapids-package-name` might be good for us in general in the long run, there's more implicit magic there than I really like and it makes naming hard to control. That said, it wouldn't really help with the `*-cu12` wheel naming and I don't think we want to change that, so it may not ever be possible to enable what we were originally hoping for in this PR.

PR #168, opened 2024-01-04
Create Project Automation Shared Workflows

### LINKED ISSUE #163: GitHub Project Automation Shared Workflows Approach
# Project Automation - Overview
We use GitHub issues, discussions, and projects to a varying degree for RAPIDS project management. To make the Project experience better, we can automate field values such as status or release based on events occuring to the item. As an example, we can synchronize the status between PRs and their linked issues when an update occurs to the PR automatically.

I'm looking for any and all feedback to see if this approach seems viable. I plan on being the lead to implement/maintain these, but it would be good to get another person who could review them as well so I know that adds some cost to our `shared-workflows` team.

#### Where We Stand Today
I've written [a number of automations](https://github.com/NVIDIA/cccl/tree/main/.github/workflows) for the CCCL team, but they're hard to read/understand, and hard to maintain as well. When thinking about rolling out project automations across RAPIDS, I don't think it's feasible to use a bespoke approach. Enter reusable workflows.

#### Proposed Approach
Not all teams are going to want the same automations, so creating reusable workflows that perform specific tasks isn't my ideal approach. Instead, I'd like to create a number of very basic workflows that we can piece together to create complex project automations.

```
.github/workflows 
├── project-get-item-id
├── project-get-linked-issues
├── project-get-set-single-select-field
├── project-get-set-iteration-field
├── project-get-set-date-field
├── project-get-set-text-numeric-field
```

This would enable workflows similar to below:
```yml
name: Project - Set Sprint

on:
  pull_request_target:
    # We will ensure the sprint is the current sprint when an issue is (re)opened, synchronized, or edited
    types: [opened, synchronize, reopened, edited]

jobs:
  get-pr-project-id:
    permissions:
        contents: read
    uses: rapidsai/shared-workflows/.github/workflows/project-get-item-id.yml@main
    with:
        PROJECT_ID: "PVT_abcd123"
    secrets: inherit
    
  get-and-set-current-sprint-option-id:
    permissions:
          contents: read
    uses: rapidsai/shared-workflows/.github/workflows/project-get-set-iteration-option-id.yml@main
    with:
        ITEM_ID: ${{ needs.get-pr-project-id.outputs.ITEM_ID }}
        FIELD_ID: "PVTIF_abcd123"
        
        # true performs the mutation to update the item as well as outputting the option id
        # default = false
        UPDATE-ITEM: true  
        
        # true performs a secondary reusable workflow: project-get-linked-issues and updates them alongside the PR
        # default = false
        UPDATE-LINKED-ISSUES: true
  secrets: inherit
```

### Reasoning
My intention is that a team can say "I want this field to update on that trigger" and write the automation in an almost block-code way. I could envision this as a GUI similar to Lego Mindstorms/labVIEW almost. It also significantly reduces the amount of work to maintain graphQL api changes, removing any graphQL api calls from our repos except for `shared-workflows`.

This approach will create at least 6 new workflows in this repo, and I could imagine more popping up as either new graphQL/project functionality exists, or a new useage I'm not considering approaches. It would be nice if we could have subdirectories of Actions as I do not want to overrun the folder.

jarmak-nv:
Closes #163

This PR creates 5 new reusable workflows, all for project automation tasks.
- #### `project-get-item-id.yaml`
   - All project automation workflows need to call this once, projects use a custom ID for items within them rather than the global node id of the item. This returns the ID needed for downstream automations
- #### `project-get-set-iteration-field.yaml`
   - This finds the current iteration and returns it
   - if `UPDATE_ITEM` is set to `true`, it sets the field for the item to the current iteration as well
   - if `UPDATE_LINKED_ISSUES` is set to `true`, it also sets the same iteration to linked issues
- #### `project-get-set-single-select-field.yaml`
   - This finds a specific `OptionId` for a provided single select value and returns it
   - if `UPDATE_ITEM` is set to `true`, it sets the field for the item to the given value as well
   - if `UPDATE_LINKED_ISSUES` is set to `true`, it also sets the same field to linked issues
- #### `project-set-text-date-numeric-field.yaml`
   - These three all follow similar usage patterns in graphQL so they're combined
   - Currently, these only set values, but if given a necessary usecase we could switch to a `get-set` as well
   - if `UPDATE_LINKED_ISSUES` is set to `true`, it also sets the same field to linked issues
- #### `project-update-linked-issues.yaml`
   - The `get-set/set` workflows all call this workflow when `UPDATE_LINKED_ISSUES` is `true` 

These workflows have all been tested:
- [On this PR ](https://github.com/jarmak-nv/workflow-run-testing/pull/3)
- [Successful run link is here](https://github.com/jarmak-nv/workflow-run-testing/actions/runs/7412033921?pr=3) 
- [Caller Workflow is here](https://github.com/jarmak-nv/workflow-run-testing/blob/main/.github/workflows/Test_Workflows.yml)
   - You may note how there are hard-coded graphQL IDs for project fields in the caller workflow
      - This is done because otherwise we need to constantly re-query the fields and I want to minimize querying to prevent any possible graphQL API rate limiting against the token. Per GitHub these are static values so we shouldn't get caught out by changing values.

### TODO BEFORE MERGE
Currently all of the reusable workflows have the secret set to `PROJECT_MANAGEMENT_SECRET`, this needs to be changed to `ADD_TO_PROJECT_GITHUB_TOKEN` which is a token scoped specifically for project-related tasks.

### DIFF
--- a/.github/workflows/project-get-item-id.yaml
+++ b/.github/workflows/project-get-item-id.yaml
@@ -73,7 +73,7 @@ jobs:
             
             # Use jq to do the actual filtering
             item_project_id=$(jq -r '.data.node.projectItems.nodes[] |
-                           select(.project.id == "${{ inputs.PROJECT_ID }}") |
-                           .id' project_data.json)
+                              select(.project.id == "${{ inputs.PROJECT_ID }}") |
+                              .id' project_data.json)
             echo "ITEM_PROJECT_ID=$item_project_id" >> $GITHUB_OUTPUT
         continue-on-error: true
--- a/.github/workflows/project-get-item-id.yaml
+++ b/.github/workflows/project-get-item-id.yaml
@@ -27,8 +27,6 @@ on:
 
 jobs:
   get_items_project_id:
-    permissions:
-      contents: read
     runs-on: ubuntu-latest
     outputs:
       ITEM_PROJECT_ID: ${{ steps.get_item_id.outputs.ITEM_PROJECT_ID }}
@@ -77,5 +75,6 @@ jobs:
             item_project_id=$(jq -r '.data.node.projectItems.nodes[] |
                            select(.project.id == "${{ inputs.PROJECT_ID }}") |
                            .id' project_data.json)
-            echo "ITEM_PROJECT_ID=$item_project_id" >> "$GITHUB_OUTPUT"
+            cat project_data.json
+            echo "ITEM_PROJECT_ID=$item_project_id" >> $GITHUB_OUTPUT
         continue-on-error: true
--- a/.github/workflows/project-get-item-id.yaml
+++ b/.github/workflows/project-get-item-id.yaml
@@ -1,0 +1,81 @@
+name: Project - Get Item ID Within the Project
+# This workflow gets the project-specific ID for an item within a project
+# All downstream queries and mutations of fields within the project require this ID
+
+on:
+  workflow_call:
+    inputs:
+      PROJECT_ID:
+        description: "The Project's graphQL node ID"
+        type: string
+        required: true
+
+      ITEM_NODE_ID:
+        description: "The issue or PR's graphQL node ID"
+        type: string
+        required: true
+
+    secrets:
+      PROJECT_MANAGEMENT_SECRET:
+        description: "Project Access Token"
+        required: true
+
+    outputs:
+      ITEM_PROJECT_ID:
+        description: "The item's project-specific ID"
+        value: ${{ jobs.get_items_project_id.outputs.ITEM_PROJECT_ID }}
+
+jobs:
+  get_items_project_id:
+    permissions:
+      contents: read
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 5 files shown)


bdice:
I might be missing something in the terminology here. What is an iteration field?

jarmak-nv:
> I might be missing something in the terminology here. What is an iteration field?

[Docs on iteration fields
](https://docs.github.com/en/issues/planning-and-tracking-with-projects/understanding-fields/about-iteration-fields)
But regardless of docs, the most common iteration field is a sprint field.  They can also represent releases, or really any block of time between two dates depending on the level of maintenance people want to apply to the field since in the project you can modify the start/end dates beyond the automatically provided "every X weeks"

ajschmidt8 (approved):
LGTM.

Btw, you can also use the [github-script action](https://github.com/actions/github-script) for making GitHub API calls.

github-script is able to make GraphQL calls and also has the benefit of supporting retries to avoid random transient networking issues:

- https://github.com/actions/github-script?tab=readme-ov-file#run-custom-graphql-queries
- https://github.com/actions/github-script?tab=readme-ov-file#retries

PR #169, opened 2024-01-04
run wheels-publish jobs in Python 3.11 image

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/7.

Split off from #166.

This proposes upgrading the image used in the `wheels-publish` workflow to a newer Python version (`3.10 -> 3.11`).

### Benefits of this change

Delays similar updates further into the future (e.g. `aws` and `anaconda-client` will drop Python 3.10 support before they drop Python 3.11 support).

### How I tested this

This workflow just runs this script: https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-wheels-anaconda.

So just checked that the things that script needs are in the image.

```shell
docker run --rm -it rapidsai/ci-wheel:cuda12.0.1-centos7-py3.11 bash

which rapids-wheels-anaconda
# /usr/local/bin/rapids-wheels-anaconda

aws --version
# aws-cli/2.15.7 Python/3.11.6 Linux/5.4.0-148-generic docker/x86_64.centos.7 prompt/off

twine --version
# twine version 4.0.2
# (importlib-metadata: 7.0.1, keyring: 24.3.0, pkginfo: 1.9.6, requests: 2.31.0, requests-toolbelt:
# 1.0.0, urllib3: 2.1.0)

anaconda --version
# anaconda Command line client (version 1.12.2)
```

I think that's probably sufficient. It'd be easy to revert this if it breaks something... less effort and lower-risk than trying to test on another repo where this `wheels-publish` only runs on merges to `main`.

### DIFF
--- a/.github/workflows/wheels-publish.yaml
+++ b/.github/workflows/wheels-publish.yaml
@@ -41,8 +41,8 @@ jobs:
     name: wheels publish
     runs-on: linux-amd64-cpu4
     container:
-      # ctk version of the container is irrelevant in the publish step
-      # it's simply a launcher for twine
+      # CUDA toolkit version of the container is irrelevant in the publish step.
+      # This just uploads already-built wheels to remote storage.
       image: "rapidsai/ci-wheel:cuda12.2.2-centos7-py3.11"
       env:
         RAPIDS_BUILD_TYPE: ${{ inputs.build_type }}
--- a/.github/workflows/wheels-publish.yaml
+++ b/.github/workflows/wheels-publish.yaml
@@ -43,7 +43,7 @@ jobs:
     container:
       # CUDA toolkit version of the container is irrelevant in the publish step.
       # This just uploads already-built wheels to remote storage.
-      image: "rapidsai/ci-wheel:cuda12.2.2-centos7-py3.11"
+      image: "rapidsai/ci-wheel:cuda12.0.1-centos7-py3.11"
       env:
         RAPIDS_BUILD_TYPE: ${{ inputs.build_type }}
     steps:
--- a/.github/workflows/wheels-publish.yaml
+++ b/.github/workflows/wheels-publish.yaml
@@ -43,7 +43,7 @@ jobs:
     container:
       # ctk version of the container is irrelevant in the publish step
       # it's simply a launcher for twine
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels-publish.yaml
```diff
@@ -41,9 +41,9 @@ jobs:
     name: wheels publish
     runs-on: linux-amd64-cpu4
     container:
-      # ctk version of the container is irrelevant in the publish step
-      # it's simply a launcher for twine
-      image: "rapidsai/ci-wheel:cuda12.0.1-centos7-py3.10"
+      # CUDA toolkit version of the container is irrelevant in the publish step.
+      # This just uploads already-built wheels to remote storage.
```
jameslamb, line 45: This workflow no longer uses `twine`, as of #154.

Proposing changing this comment to something correct and more future-proof.

bdice (approved):
I'm okay with this change but I would consider delaying on merging this until we begin work (or complete work) on https://github.com/rapidsai/build-planning/issues/3, currently targeting release 24.04. It doesn't benefit us very much to jump ahead with a new Python version until it's widely used across RAPIDS. Moving all of RAPIDS in lockstep is my preference.

jameslamb:
Ok makes sense, thanks for that! I agree, this should only be merged as part of https://github.com/rapidsai/build-planning/issues/3, in whatever release that targets.

bdice:
I converted this to a draft for now. We can reopen once we're moving forward on Python 3.11.

vyasr:
The Python version used in this jobs can safely be decoupled from everything else, but conceptually I'm fine with keeping things in sync. No rush here.

jakirkham:
As we are now adding Python 3.11, marking ready for review

bdice (approved):

bdice:
I created a test PR for cudf here: https://github.com/rapidsai/cudf/pull/15227

If that passes, we can close it and merge this PR.

bdice:
I realized that the `wheels-publish` workflow only runs on `build.yaml`, not `pr.yaml`. So https://github.com/rapidsai/cudf/pull/15227 isn't a sufficient test. There isn't a good way to test this because it only triggers when PRs are merged (or during nightly tests). I'm going to go ahead and merge it so we can get feedback from nightlies and revert tomorrow if necessary.

PR #171, opened 2024-01-18
Remove default exclusions from symbol_exclusions

robertmaynard:
Needed as we know have patches for CCCL ( cub, thrust, libcudacxx ) which hide all symbols. So projects should be able to run without any exclusions at all.

### DIFF
--- a/.github/workflows/conda-cpp-post-build-checks.yaml
+++ b/.github/workflows/conda-cpp-post-build-checks.yaml
@@ -85,9 +85,11 @@ jobs:
           path: "./tool/"
           fetch-depth: 0
       - name: Verify CUDA libraries have no public kernel entry points
+        env:
+          SYMBOL_EXCLUSIONS: ${{ inputs.symbol_exclusions }}
         run: |
-          if ${{ inputs.symbol_exclusions }}; then
-            python ./tool/detect.py ${RAPIDS_EXTRACTED_DIR}/lib -e "${{ inputs.symbol_exclusions }}"
+          if [ -n "${SYMBOL_EXCLUSIONS}" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/conda-cpp-post-build-checks.yaml
```diff
@@ -86,4 +85,9 @@ jobs:
           path: "./tool/"
           fetch-depth: 0
       - name: Verify CUDA libraries have no public kernel entry points
-        run: python ./tool/detect.py ${RAPIDS_EXTRACTED_DIR}/lib -e "${{ inputs.symbol_exclusions }}"
+        run: |
+          if ${{ inputs.symbol_exclusions }}; then
+            python ./tool/detect.py ${RAPIDS_EXTRACTED_DIR}/lib -e "${{ inputs.symbol_exclusions }}"
+          else
+            python ./tool/detect.py ${RAPIDS_EXTRACTED_DIR}/lib
+          fi
```
bdice, line 93:
```suggestion
        env:
          SYMBOL_EXCLUSIONS: ${{ inputs.symbol_exclusions }}
        run: |
          if [ -n "${SYMBOL_EXCLUSIONS}" ]; then
            python ./tool/detect.py ${RAPIDS_EXTRACTED_DIR}/lib -e "${SYMBOL_EXCLUSIONS}"
          else
            python ./tool/detect.py ${RAPIDS_EXTRACTED_DIR}/lib
          fi
```

robertmaynard:
https://github.com/rapidsai/cudf/actions/runs/7572156073/job/20628744501 shows the new script running as expected

bdice (approved):

PR #173, opened 2024-01-23
Add test summaries to wheel jobs

bdice:
This PR adds test summaries to wheel jobs. Supports https://github.com/rapidsai/cudf/pull/14847 for cudf.pandas test summaries.

### DIFF
--- a/.github/workflows/wheels-test.yaml
+++ b/.github/workflows/wheels-test.yaml
@@ -136,6 +136,12 @@ jobs:
       env:
         GH_TOKEN: ${{ github.token }}
 
+    - name: Generate test report
+      uses: test-summary/action@v2
+      with:
+        paths: "${{ env.RAPIDS_TESTS_DIR }}/*.xml"
+      if: always()
+
     - name: Upload additional artifacts
       if: "!cancelled()"
       run: rapids-upload-artifacts-dir cuda${RAPIDS_CUDA_VERSION%%.*}_$(arch)_py${RAPIDS_PY_VERSION//.}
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels-test.yaml
```diff
@@ -24,7 +24,11 @@ on:
       container-options:
         required: false
         type: string
-        default: '-e _NOOP'
+        default: "-e _NOOP"
+      test_summary_show:
```
bdice, line 28: By default, the `test-summary` action shows output from each of the failures. Since we have a large absolute number of failures in the cudf.pandas test suite (overall it's like 90%+ passing but there are a ton of tests), the output exceeds the GitHub summary size limit. I needed to add this so that we could hide the summary like this:

```
# Hide test failures because they exceed the GITHUB_STEP_SUMMARY output limit.
test_summary_show: "none"
```

Should I add this `test_summary_show` option to `.github/workflows/conda-cpp-tests.yaml` or `.github/workflows/conda-python-tests.yaml`, too? Or just leave it on the wheel jobs until it is needed elsewhere?

ajschmidt8: >Should I add this test_summary_show option to .github/workflows/conda-cpp-tests.yaml or .github/workflows/conda-python-tests.yaml, too? Or just leave it on the wheel jobs until it is needed elsewhere?

Up to you. I don't expect this option to be widely used, so I'm fine with omitting it from the other workflows until it's needed.

ajschmidt8 (approved):

PR #176, opened 2024-02-20
add support for python 3.11

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/3.

Proposes adding Python 3.11 entries in the following workflows:

* `conda-python-build`
* `conda-python-test`
* `wheels-build`
* `wheels-test`

### Why these specific matrix entries?

Following the constraints laid out in the description of https://github.com/rapidsai/build-planning/issues/3, and the discussion from #166, I'm proposing:

* adding Python 3.11 *builds* for every combination where we currently do 3.11 builds (since those don't require GPUs)
* replacing some Python 3.10 *tests* with Python 3.11
   - holding the absolute number of testing jobs unchanged
   - trying to spread Python 3.11 test jobs across CUDA version, operating system, and architecture
   - trying to cover "oldest of everything"
   - trying to cover "latest of everything"

### Rollout plan?

* [ ] 1. merge this PR

I'm proposing using GitHub's `continue-on-error` ([docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error)) to allow the Python 3.11 jobs to fail on PRs without blocking those PRs from being merged.

So this could be merged into `branch-24.04` without making updates in all the repos.

* [ ] 2. update all the repos, in RAPIDS dependency order

Using something like https://github.com/rapidsai/rapids-reviser/pull/11, go update all the repos... adding Python 3.11 blocks to `dependencies.yaml`, updating `pyproject.toml` and docs, and getting these CI jobs all passing.

* [ ] 3. make these jobs required

Once all RAPIDS repos have been updated and these. Python 3.11 jobs are passing for them, making these jobs required to prevent further development from breaking Python 3.11 support

* [ ] 4. other steps in "Post Migration" described at https://github.com/rapidsai/build-planning/issues/3
* [ ] 5. switch to Python 3.11 in `wheels-publish` CI job
   - [ ] #169

## How I tested this

Put up https://github.com/rapidsai/rmm/pull/1469, with extra changes pointing its GitHub Actions configuration at this branch. Saw all the Python 3.11 jobs expected get run (and they happened to run successfully, yay!). Looked through the logs and confirmed that Python 3.11 was really being pulled in there, and that wheels had names like `rm-*-cp311-*.whl`.

Also put up https://github.com/rapidsai/cudf/pull/15111, to test a more complex pipeline. Saw all the Python 3.11 jobs expected get run... many of them failed, but for the expected types of reasons... for example, there weren't `rmm-cu12==24.4.*` wheels supporting Python 3.11 available yet ([build link](https://github.com/rapidsai/cudf/actions/runs/8007612080/job/21872241250?pr=15111#step:9:60)).

But to be clear, I only changed those GitHub Actions configs to point to this branch **to test this PR**. That shouldn't be necessary for any of the other PRs, if we go with the `continue-on-error` approach I'm proposing here.

### DIFF
--- a/.github/workflows/conda-python-build.yaml
+++ b/.github/workflows/conda-python-build.yaml
@@ -77,6 +77,7 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
     runs-on: "linux-${{ matrix.ARCH }}-${{ inputs.node_type }}"
+    # TODO: remove 'continue-on-error' before closing https://github.com/rapidsai/build-planning/issues/3
     continue-on-error: ${{ matrix.PY_VER == '3.11' }}
     env:
       RAPIDS_ARTIFACTS_DIR: ${{ github.workspace }}/artifacts
--- a/.github/workflows/conda-python-build.yaml
+++ b/.github/workflows/conda-python-build.yaml
@@ -57,10 +57,14 @@ jobs:
           - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.9' }
           - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10' }
           - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10' }
+          - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.11' }
+          - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.11' }
           - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.9' }
           - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.9' }
           - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10' }
           - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10' }
+          - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.11' }
+          - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.11' }
           "
 
           echo "MATRIX=$(
@@ -73,6 +77,7 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
     runs-on: "linux-${{ matrix.ARCH }}-${{ inputs.node_type }}"
+    continue-on-error: ${{ matrix.PY_VER == '3.11' }}
     env:
       RAPIDS_ARTIFACTS_DIR: ${{ github.workspace }}/artifacts
     container:
--- a/.github/workflows/conda-python-build.yaml
+++ b/.github/workflows/conda-python-build.yaml
@@ -62,11 +62,15 @@ jobs:
           - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04' }
           - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04' }
           - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04' }
+          - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04' }
+          - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04' }
           # arm64
           - { ARCH: 'arm64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04' }
           - { ARCH: 'arm64', PY_VER: '3.9',  CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04' }
           - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04' }
           - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04' }
+          - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04' }
+          - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04' }
           "
 
           echo "MATRIX=$(
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -95,6 +95,7 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
     runs-on: "linux-${{ matrix.ARCH }}-gpu-${{ matrix.GPU }}-${{ matrix.DRIVER }}-1"
+    # TODO: remove 'continue-on-error' before closing https://github.com/rapidsai/build-planning/issues/3
     continue-on-error: ${{ matrix.PY_VER == '3.11' }}
     env:
       RAPIDS_COVERAGE_DIR: ${{ github.workspace }}/coverage-results
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -80,9 +80,9 @@ jobs:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest' }
               - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', GPU: 'v100', DRIVER: 'latest'   }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'latest'   }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
-              - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
+              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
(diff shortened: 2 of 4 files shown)


.github/workflows/conda-python-build.yaml
```diff
@@ -73,6 +77,7 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
     runs-on: "linux-${{ matrix.ARCH }}-${{ inputs.node_type }}"
+    continue-on-error: ${{ matrix.PY_VER == '3.11' }}
```
KyleFromNVIDIA, line 80: What is the purpose of these `continue-on-error` directives?

jameslamb: I'll explain this note from the description in more detail:

> *I'm proposing using GitHub's continue-on-error ([docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error)) to allow the Python 3.11 jobs to fail on PRs without blocking those PRs from being merged.*

Doing this allows us to merge this PR, add support one RAPIDS repo at a time, and then come back and remove this to make those required again.

If we just merged this PR as-is without `continue-on-error`, CI would be broken in any RAPIDS repos not yet supporting Python 3.11 (which is most of them right now) until they were fixed. That'd be really disruptive to development there.

Another alternative is to not do this `continue-on-error` thing, and instead do 2 rounds of PRs to all the RAPIDS repos:

* round 1: updating all their GitHub Actions configs to like `{workflow}.yaml@python-3.11`, getting all those building successfully, then merging them all in quick succession
* round 2: merging this PR, then updating all those configs in the repos again to point at `{workflow}.yaml@branch-24.04`

That's what we did for CUDA 12.2 support (https://github.com/rapidsai/build-planning/issues/7) and it increased the amount of effort required for everyone involved. Hoping to avoid that here to make this update go smoother and faster.

KyleFromNVIDIA: Sounds like a good plan!

KyleFromNVIDIA: That being said, might I suggest adding a TODO comment referencing the 3.11 issue? Something people can reference to understand why this directive is here.

KyleFromNVIDIA (changes requested):
Looks good overall, would like to see TODO comments explaining the `continue-on-error` directives.

.github/workflows/conda-python-build.yaml
jameslamb: I'm really not a fan of TODO comments in code, I've found that they tend to be forgotten more easily than issues in a backlog. There's already a checklist at the top of this PR that includes a step "make these jobs required", could that be enough?

KyleFromNVIDIA: I absolutely agree we should have the issue in a backlog. I was thinking we could *also* have a TODO comment that references said issue. WDYT?

jameslamb: Ok sure, no problem.

Added in https://github.com/rapidsai/shared-workflows/pull/176/commits/c32c724b8f2ed0b29e9b73c7661262094862456c

AyodeAwe (approved):

.github/workflows/conda-python-build.yaml
KyleFromNVIDIA: I see the comment in `wheels-build.yaml` and `wheels-test.yaml`. Please add it to `conda-python-build.yaml` and `conda-python-tests.yaml` too.

msarahan: https://blog.stackademic.com/vscode-auto-save-22b6ce1c81e1

KyleFromNVIDIA (approved):

.github/workflows/conda-python-build.yaml
jameslamb: thanks @msarahan ! added!

msarahan (approved):
> adding Python 3.11 builds for every combination where we currently do 3.11 builds (since those don't require GPUs)
I'm pretty sure you mean add 3.11 everywhere that we currently have 3.10, but just want to make certain.

I don't like the approach of keeping the same number of tests, but just assigning different combinations. I would like it better if we had random sampling without replacement. I googled around for that quickly, and it seems possible but not worth the effort. The approach here is probably good enough, even if it feels a little off.

.github/workflows/wheels-test.yaml
```diff
@@ -71,19 +71,19 @@ jobs:
...
               - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
             nightly:
               - { ARCH: 'amd64', PY_VER: '3.9', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
               - { ARCH: 'arm64', PY_VER: '3.9', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
-              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
+              - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
```
bdice, line 80: Let’s reshuffle these a bit. I want one configuration to be amd64 + latest Python + latest CUDA. Currently it seems only ARM has Python 3.11 jobs. I would also tilt the distribution of versions to be evenly weighted, it seems like the matrix tests 3.9 and 3.10 more than 3.11 (I didn’t do an exact count).

jameslamb: Thanks for looking @bdice .

This PR *does* currently have a Python 3.11 AMD job for `wheels-test`... but just on PRs.

```text
- { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu20.04', gpu: 'v100', driver: 'latest' }
```

Definitely agree the mix across Python versions is uneven right now, I'll fix that.

Before I go further with this... would you consider this change I just put up? 

https://github.com/rapidsai/shared-workflows/pull/177

Based on my experience here and in #166, I think it's worth taking a minute to re-sorting these matrices so they're easier to review.

jameslamb: Alright @bdice whenever you have time, could you please take another look?

(and @msarahan @AyodeAwe @KyleFromNVIDIA if you'd like another chance to review now that changes have been made)

By my count, on the current state of this branch we have the following mix:

| workflow                            | python version | amd64 | arm64 | total |
|:-------------------------:|:--------------:|:--------:|:------:|:-----:|
| conda-python-build         | 3.9                    |  2           | 2         | 4       |
| conda-python-build         | 3.10                    |  2           | 2         | 4       |
| conda-python-build         | 3.11                    |  2           | 2         | 4       |
| conda-python-tests (pr) | 3.9                     | 1            | 0          | 1        | 
| conda-python-tests (pr) | 3.10                   | 1             | 1         | 2        |
| conda-python-tests (pr) | 3.11                   | 1             | 0         | 1         |
| conda-python-tests (nightly) | 3.9            | 2            | 1          | 3        | 
| conda-python-tests (nightly) | 3.10           | 2             | 1         | 3        |
| conda-python-tests (nightly) | 3.11            | 1             | 2        | 3        |
| wheel-build         | 3.9                    |  4           | 2         | 6       |
| wheel-build         | 3.10                    |  4           | 2         | 6       |
| wheel-build         | 3.11                    |  4           | 2         | 6       |
| wheel-test (pr) | 3.9                     | 1            | 0          | 1        | 
| wheel-test (pr) | 3.10                   | 0             | 1         | 1        |
| wheel-test (pr) | 3.11                   | 1             | 0         | 1         |
| wheel-test (nightly) | 3.9            | 2            | 1          | 3        | 
| wheel-test (nightly) | 3.10           | 2             | 2         | 4        |
| wheel-test (nightly) | 3.11            | 2             | 1        | 3        |

And the `conda-python-tests (pr)` and `wheel-test (pr)` both have the combination "amd64 + latest Python + latest CUDA".

I've still preserved the same absolute number of *test* jobs as `branch-24.04`.

I've also updated the description, with tests from these two PRs:

* https://github.com/rapidsai/rmm/pull/1469
* https://github.com/rapidsai/cudf/pull/15111

.github/workflows/conda-python-tests.yaml
```diff
@@ -73,21 +73,21 @@ jobs:
...
               # arm64
               - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest'   }
             nightly:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest' }
               - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', GPU: 'v100', DRIVER: 'latest'   }
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'latest'   }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
+              - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
```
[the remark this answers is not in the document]
jameslamb: Promise I didn't ignore your previous comment asking for that combination.

On the current state of this branch, that exact combination is being tested *on PRs* for `conda-python-tests` workflow, which is why I didn't also add it to the nightly matrix.

Does that change you opinion on this suggestion?

Or do you think it's desirable to have that combination in both places because we don't enforce that PRs have to be up to date with the target branch before merging?

vyasr:
> > adding Python 3.11 builds for every combination where we currently do 3.11 builds (since those don't require GPUs)
> > I'm pretty sure you mean add 3.11 everywhere that we currently have 3.10, but just want to make certain.
> 
> I don't like the approach of keeping the same number of tests, but just assigning different combinations. I would like it better if we had random sampling without replacement. I googled around for that quickly, and it seems possible but not worth the effort. The approach here is probably good enough, even if it feels a little off.

I'd say we stick with the approach in this PR for now. Once this PR merges, I plan to seriously evaluate changes to address rapidsai/build-planning#5

vyasr (approved):
I'm comfortable with the choices currently made in this PR for now. Once this merges, I'd like to focus on addressing rapidsai/build-planning#5 in a PR where we can relax constraints around the number of test jobs (I agree with conserving that number in this PR to avoid doing too many things at once).

.github/workflows/conda-python-tests.yaml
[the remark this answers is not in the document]
jameslamb: Just pushed https://github.com/rapidsai/shared-workflows/pull/176/commits/115ad5aa03616c5e3464da7e65b0efd113059871. As of that commit, I believe this branch is meeting these constraints for test job matrices:

* at least 1 PR job for every Python version in `(3.9, 3.10, 3.11)`
* at least 1 PR job for every CUDA version in `(11.8.0, 12.0.1, 12.2.2)`
* at least 1 PR job for every architecture in `(amd64, arm64)`
* am64 + latest CUDA + latest Python + latest Ubuntu present in both PR and nightly
* nightly is a strict superset of what's tested on PRs
* no net change to the number of test jobs on PRs or nightly, relative to the current state of `branch-24.04`
* as-close-as-possible-to-equal spread of test jobs across Python versions

Since that commit only changed `CUDA_VER` and `LINUX_VER`, the table from https://github.com/rapidsai/shared-workflows/pull/176#discussion_r1499601279 is unchanged.

bdice (approved):
Thanks for the many iterations on this. For future work on this matrix, I am open to discussion on the rules we follow. I recognize that I asked for several iterations for various requests but I also acknowledge that some of those requests are opinionated but not necessarily right or wrong.

AyodeAwe:
Merging on author's request

Issue #118, opened 2023-07-20
Address workflow inconsistencies


ajschmidt8:
There are a lot of inconsistencies in this repository that we should work to address as time permits.

A few that come to mind are:

- workflows have both `.yml` and `.yaml` extensions
- workflows matrices use different parameter names (e.g. `CUDA_VER` for conda vs. `ctk` for wheels)
- workflows use different methods for generating timestamps (wheels use a `wheel-epoch-timestamp` job, whereas conda workflows use a `gha-tools` script that uses `gh` to retrieve the workflow start time)
- the `build` job names in conda and wheel workflows are different (`build` for conda and `wheel-build` for wheels)
  - the `name` property in these `build` jobs are also inconsistent (it's unset for conda workflows, but set for wheel workflows). this leads to visual differences when viewing the workflows in the browser (see screenshot 1 below)

This is a non-exhaustive list.

We should review the workflows for additional differences.

<details><summary>screenshot 1</summary>
<p>

![image](https://github.com/rapidsai/shared-action-workflows/assets/7400326/72e87e80-73ec-4c67-908e-4a2afce4cff5)

</p>
</details>

jameslamb:
> the name property in these build jobs are also inconsistent (it's unset for conda workflows, but set for wheel workflows)

I just discovered this (forgot about this issue) over in #177.

Not having `name:` set meant that changes that I made to the matrices for some jobs were a bit hard to compare to previous runs, because the names were in a different format.

I think setting `name:` for all these workflows would be nice... keeping that formatting consistent would allow people developing on RAPIDS repos to quickly scan the list of CI checks at a glance and identify the ones they're looking for.

Something like this:

```text
  tests:
    needs: compute-matrix
    strategy:
      fail-fast: false
      matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
    name: wheel-build (${{ matrix.CUDA_VER }}, ${{ matrix.PY_VER }}, ${{ matrix.ARCH }}, ${{ matrix.LINUX_VER }})
```

(exact details TBD, based on what's in the matrix for the particular workflow)

I *think* that should be a minimally-invasive changes, since as far as I know, we don't use those names in branch protections, but instead use this mechanism:

https://github.com/rapidsai/shared-workflows/blob/da626b622bd9cfcdd24a1968979ef80f0581b1be/.github/workflows/pr-builder.yaml#L1-L9

AyodeAwe:
[...]

PR #177, opened 2024-02-21
re-organize build matrices

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/3

Related to https://github.com/rapidsai/build-planning/issues/5

Proposes re-organizing the build matrices.

Specifically:

* using the same order of keys within each matrix entry, across different workflow configurations
* sorting all matrices in ascending order by the same keys
* adding extra whitespace so all entries for the same key are vertically aligned
* using comments to visually distinguish between `amd64` and `arm64` jobs
  - *(extra tough to see visually because those 2 strings are so similar and so short)*

### Benefits of these changes

I believe the proposed changes here will make changes like "add a new CUDA version" or "drop support for a Python version" easier to author and review.

Inspired by 2 recent experiences I've had where I proposed changes to the build matrices and both I and reviewers found them difficult to assess:

* #176
* #166 

I hope this re-organization will make it easier to determine at a glance whether matrices are achieving goals like:

- *"good balance across Python versions"*
- *"covering all CUDA `{major}.{minor}` RAPIDS wants to support"*
- *"at least one job with latest Python, latest CUDA, latest operating system"*

## Notes for Reviewers

To be clear, **this is is only a moving-code-around PR**.

It does not propose any changes to the build matrices for any workflows. If you see any places where this PR would lead to adding a new type of job or dropping a job that's currently being run, that's a mistake.

## How I tested this

Opened https://github.com/rapidsai/cudf/pull/15111 to test.

The matrix of jobs there (https://github.com/rapidsai/cudf/actions/runs/7994808681/job/21834408169?pr=15111) looks identical to the most recent successful run from some other PR targeting `branch-24.04` (https://github.com/rapidsai/cudf/actions/runs/7992742449).

Also spot-checked the logs to confirm that we were getting the right runner architecture, Python version, CUDA version, and operating system.

Some of the job *names* did change... I think because we don't specify `name:` explicitly in the workflows here, so they're auto-generated based on the matrix entries.

image

*(left = before, right = as of this PR)*

I don't think that should affect the correctness of branch protections, since we have this mechanism:

https://github.com/rapidsai/shared-workflows/blob/da626b622bd9cfcdd24a1968979ef80f0581b1be/.github/workflows/pr-builder.yaml#L1-L9

That's discussed over in https://github.com/rapidsai/shared-workflows/issues/118.

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -52,11 +52,17 @@ jobs:
         run: |
           set -eo pipefail
 
+          # please keep the matrices sorted in ascending order by the following:
+          #
+          #     [ARCH, PY_VER, CUDA_VER, LINUX_VER]
+          #
           export MATRIX="
-          - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10' }
-          - { CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10' }
-          - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'amd64', PY_VER: '3.10' }
-          - { CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', ARCH: 'arm64', PY_VER: '3.10' }
+          # amd64
+          - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04' }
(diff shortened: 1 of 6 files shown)


msarahan (approved):

ajschmidt8 (approved):

msarahan:
image

jameslamb:
haha thanks @msarahan .

@ajschmidt8 could you merge this?

msarahan:
let's see if I have any power here...

/merge

bdice:
image

image

jameslamb:
Thanks @bdice !

bdice:
I don’t think this repo uses /merge commands?

Also FYI, the comment must be only /merge and not a message containing /merge. @msarahan’s attempt would not have worked because it had content before the /merge.

jameslamb:
Thanks for that! For the record, I didn't have the green `merge` button available, so either way I suspect I don't have sufficient access to merge here.

PR #182, opened 2024-02-28
always mark Python 3.11 jobs as successful

jameslamb:
Follow-up to #176.

#176 added Python 3.11 build and test jobs. It tried to make those optional by using  GitHub Actions' `jobs.<job_id>.continue_on_error` mechanism ([docs on that](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error)).

Unfortunately, it seems that that mechanism doesn't play nicely with RAPIDS branch protections, which use one more job that runs at the end of the `pr` workflow and declares all other sets of jobs in a `needs:` block.

`cudf` example: https://github.com/rapidsai/cudf/blob/8526e6d5b21361465d1c72ecbea64d3d2d9bf849/.github/workflows/pr.yaml#L13-L34.

As a result, the Python 3.11 jobs **were actually blocking CI** in repos where they were failing.

This proposes the following changes:

* removing that job-level `continue-on-error`
* ensuring that Python 3.11 jobs are always marked successful, even when individual steps in them fail

### How I tested this

Opened https://github.com/rapidsai/cudf/pull/15172 to test.

On the first run, I saw all Python 3.11 marked as successful, even though their logs clearly indicated that they failed.

Screenshot 2024-02-28 at 10 56 20 AM

([build link](https://github.com/rapidsai/cudf/actions/runs/8083300129/job/22086966999?pr=15172))

~TODO: once that completes, I'll do one more build that breaks the tests and shows that the non-Python-3.11 jobs are marked as failed.~

Proposing that we merge this without waiting on that test, to unblock CI for `cudf` (and maybe other projects).

### DIFF
--- a/.github/workflows/wheels-test.yaml
+++ b/.github/workflows/wheels-test.yaml
@@ -149,11 +149,15 @@ jobs:
         sha: ${{ inputs.sha }}
 
     - name: Run tests
+      # TODO: remove 'continue-on-error' before closing https://github.com/rapidsai/build-planning/issues/3
+      continue-on-error: ${{ matrix.PY_VER == '3.11' }}
       run: ${{ inputs.script }}
       env:
         GH_TOKEN: ${{ github.token }}
 
     - name: Generate test report
+      # TODO: remove 'continue-on-error' before closing https://github.com/rapidsai/build-planning/issues/3
+      continue-on-error: ${{ matrix.PY_VER == '3.11' }}
       uses: test-summary/action@v2.2
       with:
         paths: "${{ env.RAPIDS_TESTS_DIR }}/*.xml"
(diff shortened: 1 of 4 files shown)


.github/workflows/wheels-test.yaml
```diff
@@ -149,17 +147,29 @@ jobs:
...
       if: always()
 
     - name: Upload additional artifacts
+      # TODO: remove 'continue-on-error' before closing https://github.com/rapidsai/build-planning/issues/3
+      continue-on-error: ${{ matrix.PY_VER == '3.11' }}
       if: "!cancelled()"
       run: rapids-upload-artifacts-dir cuda${RAPIDS_CUDA_VERSION%%.*}_$(arch)_py${RAPIDS_PY_VERSION//.}
+
+    # TODO: remove before closing https://github.com/rapidsai/build-planning/issues/3
+    - name: mark Python 3.11 jobs successful
```
bdice, line 172: I'm not 100% sure if this step is necessary, but it doesn't hurt. We have enough testing to know that this PR will help unblock CI.

ajschmidt8 (approved):

bdice (approved):

PR #184, opened 2024-03-01
Reduce CI runs

vyasr:
This PR cuts down on the the number testing jobs in PRs while attempting to retain similar coverage. It also enables `fail-fast` so that matrix jobs will not proceed to completion if any one of the tests fail.

Contributes to rapidsai/build-planning#5

### DIFF
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -97,7 +97,7 @@ jobs:
   tests:
     needs: compute-matrix
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
     runs-on: "linux-${{ matrix.ARCH }}-gpu-${{ matrix.GPU }}-${{ matrix.DRIVER }}-1"
     env:
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -68,11 +68,10 @@ jobs:
           export MATRICES="
             pull-request:
               # amd64
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'centos7',      GPU: 'v100', DRIVER: 'latest'  }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04',  GPU: 'v100', DRIVER: 'latest'  }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04',  GPU: 'v100', DRIVER: 'latest'  }
+              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest'   }
+              - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
               # arm64
-              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04',  GPU: 'a100', DRIVER: 'latest'  }
+              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest'   }
             nightly:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest' }
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -71,11 +71,7 @@ jobs:
           export MATRICES="
             pull-request:
               # amd64
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest' }
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
-              - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
-              # arm64
-              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest'   }
             nightly:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest' }
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -74,7 +74,7 @@ jobs:
               - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest'   }
               - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
               # arm64
-              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest'   }
+              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest'   }
             nightly:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest' }
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -71,7 +71,10 @@ jobs:
           export MATRICES="
             pull-request:
               # amd64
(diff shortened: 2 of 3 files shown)


.github/workflows/conda-python-tests.yaml
```diff
@@ -71,11 +71,7 @@ jobs:
           export MATRICES="
             pull-request:
               # amd64
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest' }
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
```
bdice, line 74: We must test at least one conda job for each CUDA major version due to the importance of CUDA in our conda packages. Here, I expand this to two jobs and cover as much as possible across them.
```suggestion
              - { ARCH: 'arm64', PY_VER: '3.9', CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest'   }
              - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
```

vyasr: I agree, that's a good point. How about on the wheel side? Do you think it's important to keep one job per CUDA version there? I'm torn, because I think it makes sense to have coverage and there are different problems that could happen for wheels, but I also don't know if it's worth the tradeoff right now. Wheels are definitely less likely to have problems with different CUDA versions than conda.

.github/workflows/wheels-test.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
           export MATRICES="
             pull-request:
               # amd64
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'earliest' }
               - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', gpu: 'v100', driver: 'latest' }
               # arm64
               - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
```
vyasr, line 81: I don't think we've ever seen wheel+arm be the only failing job, or at least not since the initial development of wheels. In theory there are potential glibc problems. In practice we've never observed them. I'm tempted to drop this because they're 10 minute jobs taking up our most valuable resource (arm GPU nodes).

vyasr (commented):
Will also apply fail-fast: true to the test jobs once we've finished discussions on the matrix.

.github/workflows/wheels-test.yaml
bdice: > taking up our most valuable resource (arm GPU nodes).

I'm not sure how true this is from recent observation. I think the sheer number of non-ARM jobs has led to long queue times for amd64 nodes even when no queue exists for arm64. I often see arm64 jobs finish _before_ amd64 due to the shorter queues. At one point, yes, arm64 had a longer queue, but I haven't seen that lately.

bdice: I'm not sure if I understand the original comment though. Are you proposing to remove ARM wheel jobs from the PR matrix entirely? I think it's good to have at least one ARM job for wheels.

I would propose to leave an ARM + wheel job in the matrix for now. If we see ARM queue times increasing due to a change in the ratio of amd64 to ARM jobs, then we can reconsider.

vyasr:
We discussed offline and the new matrix reflects the consensus @ajschmidt8 @bdice and I came to.

vyasr:
The latest run on the cudf test PR verifies that this workflow is valid: https://github.com/rapidsai/cudf/actions/runs/8116036240?pr=15201

vyasr:
New run based on setting v100/a100 appropriately: https://github.com/rapidsai/cudf/actions/runs/8116093123?pr=15201

.github/workflows/conda-python-tests.yaml
bdice: Resolved by discussion documented here: https://github.com/rapidsai/build-planning/issues/5#issuecomment-1973775147

.github/workflows/conda-cpp-tests.yaml
```diff
@@ -97,7 +97,7 @@ jobs:
   tests:
     needs: compute-matrix
     strategy:
-      fail-fast: false
+      fail-fast: true
```
bdice, line 100: Let's add this option explicitly to the wheel jobs, where it is currently using the default value of `true`. Just want to be explicit everywhere.

.github/workflows/conda-python-tests.yaml
```diff
@@ -71,11 +71,10 @@ jobs:
           export MATRICES="
             pull-request:
               # amd64
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest' }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
+              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest'   }
               - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
               # arm64
-              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest'   }
+              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest'   }
```
bdice, line 77: Let's retain a conda job with CUDA 12.0.
```suggestion
              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'rockylinux8', GPU: 'a100', DRIVER: 'latest'   }
```

.github/workflows/conda-cpp-tests.yaml
vyasr: It's already there.

.github/workflows/conda-cpp-tests.yaml
bdice, line 1: Can we also cut the matrix in the conda C++ tests? Seems like this should resemble the conda Python matrix.

vyasr: Yup can do that.

vyasr: Done.

bdice:
Net changes to PRs:
conda C++ tests go from 4 jobs to 3.
conda Python tests go from 4 jobs to 3.
wheel tests go from 3 jobs to 2.

amd64 jobs go from 8 (3 conda C++, 3 conda Python, 2 wheel) to 5 (2 conda C++, 2 conda Python, 1 wheel)
arm64 jobs go from 3 to ... 3. 😬

Of course, we also have to account for custom jobs like docs and notebooks that also consume amd64 runners. That usually adds about 2 jobs to the amd64 runners, depending on the repository.

For this to improve CI times, we have to be waiting on amd64. That's currently true, from my observations, and I expect that reducing amd64 usage will be sufficient to improve CI queue times.

If you also consider that some of those amd64 jobs are "earliest" driver and thus come from a separate pool of runners, this is an even more significant improvement. 👍

.github/workflows/conda-cpp-tests.yaml
```diff
@@ -68,11 +68,10 @@ jobs:
           export MATRICES="
             pull-request:
               # amd64
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'centos7',      GPU: 'v100', DRIVER: 'latest'  }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04',  GPU: 'v100', DRIVER: 'latest'  }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04',  GPU: 'v100', DRIVER: 'latest'  }
+              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'centos7',     GPU: 'v100', DRIVER: 'earliest'   }
```
bdice, line 71: Ooh, this rebalances one of the C++ jobs to use the "earliest" pool of runners. That's probably a net win.

bdice (approved):

.github/workflows/wheels-test.yaml
```diff
@@ -75,10 +75,9 @@ jobs:
           export MATRICES="
             pull-request:
               # amd64
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'v100', driver: 'latest' }
-              - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', gpu: 'v100', driver: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '12.2.2', LINUX_VER: 'rockylinux8', gpu: 'v100', driver: 'latest' }
               # arm64
-              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', gpu: 'a100', driver: 'latest' }
+              - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu18.04', gpu: 'a100', driver: 'latest' }
```
ajschmidt8, line 80: This particular combo doesn't seem to have an `arm` variant: https://hub.docker.com/layers/rapidsai/citestwheel/cuda11.8.0-ubuntu18.04-py3.11/images/sha256-71f5576469cfa14dce97e3fb4d8e7aec7c75ee3f2136396d11909f075e0a84a8?context=explore

ajschmidt8 (commented):
found two invalid combinations

bdice:
https://github.com/rapidsai/shared-workflows/pull/184#discussion_r1509457946
https://github.com/rapidsai/cudf/actions/runs/8117888463/job/22191610519?pr=15201

Confirming this hypothesis about imbalanced queues for latest/earliest drivers -- for this job's C++ tests, the amd64 "latest" job is still waiting on a runner after 45 minutes, but the amd64 "earliest" job started immediately.

vyasr:
Aside from some unrelated failures, the test cudf job looks good now. @ajschmidt8 I think this is good to go when you're happy with it.

ajschmidt8 (approved):

PR #187, opened 2024-03-04
Fail if the computed matrix is empty.

bdice:
Currently the `compute-matrix` job passes even if the computed matrix is empty. Empty matrices raise an error (as shown below) so this should not be permitted.

image

This is tested in https://github.com/rapidsai/cudf/pull/15221.

### DIFF
--- a/.github/workflows/conda-python-build.yaml
+++ b/.github/workflows/conda-python-build.yaml
@@ -73,10 +73,12 @@ jobs:
           - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04' }
           "
 
-          echo "MATRIX=$(
+          MATRIX="$(
             yq -n -o json 'env(MATRIX)' | \
-            jq -c '${{ inputs.matrix_filter }} | if (. | length) > 0 then {include: .} else "Error: Empty matrix\n" | halt_error(1) end' \
-          )" | tee --append "${GITHUB_OUTPUT}"
+            jq -c '${{ inputs.matrix_filter }} | if (. | length) > 0 then {include: .} else "Error: Empty matrix\n" | halt_error(1) end'
+          )"
+
+          echo "MATRIX=${MATRIX}" | tee --append "${GITHUB_OUTPUT}"
   build:
     needs: compute-matrix
     strategy:
--- a/.github/workflows/conda-python-build.yaml
+++ b/.github/workflows/conda-python-build.yaml
@@ -75,7 +75,7 @@ jobs:
 
           echo "MATRIX=$(
             yq -n -o json 'env(MATRIX)' | \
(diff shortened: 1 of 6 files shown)


.github/workflows/conda-python-build.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
 
           echo "MATRIX=$(
             yq -n -o json 'env(MATRIX)' | \
-            jq -c '${{ inputs.matrix_filter }} | {include: .}' \
+            jq -c '${{ inputs.matrix_filter }} | {include: .} | if .include != [] then . else halt_error(1) end' \
```
ajschmidt8, line 78:
```suggestion
            jq -c '${{ inputs.matrix_filter }} | if (. | length) > 0 then {include: .} else "Error: Empty matrix\n" | halt_error(1) end' \
```

This change checks the length of the array before constructing the `{include: .}` object.

It also prints a message describing the error before exiting.

.github/workflows/conda-python-tests.yaml
```diff
@@ -94,7 +94,7 @@ jobs:
 
           echo "MATRIX=$(
             yq -n -o json 'env(TEST_MATRIX)' | \
-            jq -c '${{ inputs.matrix_filter }} | {include: .}' \
+            jq -c '${{ inputs.matrix_filter }} | {include: .} | if .include != [] then . else halt_error(1) end' \
```
ajschmidt8, line 97:
```suggestion
            jq -c '${{ inputs.matrix_filter }} | if (. | length) > 0 then {include: .} else "Error: Empty matrix\n" | halt_error(1) end' \
```

.github/workflows/wheels-build.yaml
```diff
@@ -119,11 +119,11 @@ jobs:
 
           echo "MATRIX=$(
             yq -n -o json 'env(MATRIX)' | \
-            jq -c '${{ inputs.matrix_filter }} | {include: .}' \
+            jq -c '${{ inputs.matrix_filter }} | {include: .} | if .include != [] then . else halt_error(1) end' \
```
ajschmidt8, line 122:
```suggestion
            jq -c '${{ inputs.matrix_filter }} | if (. | length) > 0 then {include: .} else "Error: Empty matrix\n" | halt_error(1) end' \
```

bdice:
@ajschmidt8 It doesn't seem like the `halt_error` is actually causing a failure. https://github.com/rapidsai/cudf/actions/runs/8143986136/job/22257171652?pr=15221

I tried this locally and `halt_error(1)` gives an exit code of 1. Do we need to split up the pipeline somehow so that we can get the exit code to fail?

bdice:
Verified in https://github.com/rapidsai/cudf/actions/runs/8144434095/job/22258595781?pr=15221

This works as desired.

PR #190, opened 2024-03-07
Revert "Add back wheels test job for Python 3.11, amd64, CUDA 12."

vyasr:
Reverts rapidsai/shared-workflows#188

To keep things running smoothly, we're going to wait to merge this until 24.06 since it relies on fixes elsewhere in RAPIDS and we want to give ourselves a bit of time to get everything sorted out.


bdice (approved):
This should be safe to merge now, I believe we've cleaned up the places that relied on this specific job being in the matrix.

PR #191, opened 2024-03-13
Initiate switch from `build_script` and `test_script` to `script`

AyodeAwe:
The script used in wheel workflows is indicated by `script`, whereas the conda workflows use `build_script` and `test_script`.

This PR initiates a switch for the conda build and test workflows (cpp, python) to use the `script`  input name. After this PR is merged, PRs will be opened to all RAPIDS repositories which call these workflows to update the inputs from `build_script` and `test_script` (if these are present in the workflows) to `script`.

The rationale is that the naming convention of `build_script` and `test_script` appear rather redundant given the name of their respective workflows (`cpp_build`, `cpp_test`, etc.)

After these PRs are merged, another PR will be opened in this repository to remove the `build_script` and `test_script` references for these shared workflows.

This PR also adds a name for the conda build job, keeping it consistent with the named wheels build job.

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -72,6 +72,7 @@ jobs:
 
           echo "MATRIX=${MATRIX}" | tee --append "${GITHUB_OUTPUT}"
   build:
+    name:  ${{ matrix.CUDA_VER }}, ${{ matrix.PY_VER }}, ${{ matrix.ARCH }}, ${{ matrix.LINUX_VER }}
     needs: compute-matrix
     timeout-minutes: 480
     strategy:
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -17,7 +17,6 @@ on:
         default: "cpu8"
       build_script:
         type: string
-        default: "ci/build_cpp.sh"
       script:
         type: string
         default: "ci/build_cpp.sh"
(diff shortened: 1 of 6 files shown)


.github/workflows/wheels-build.yaml
```diff
@@ -18,6 +18,8 @@ on:
       script:
         required: true
```
raydouglass, line 19: Should this input be switched to not required?

AyodeAwe: Yeah looks like that should be removed.

.github/workflows/conda-cpp-build.yaml
```diff
@@ -72,6 +74,7 @@ jobs:
 
           echo "MATRIX=${MATRIX}" | tee --append "${GITHUB_OUTPUT}"
   build:
+    name:  ${{ matrix.CUDA_VER }}, ${{ matrix.PY_VER }}, ${{ matrix.ARCH }}, ${{ matrix.LINUX_VER }}
```
ajschmidt8, line 77: was this change intended to be included in this PR? it's not mentioned in the PR body. I know others may have thoughts on what type of naming convention we use, so it might be best to postpone these changes for now.

AyodeAwe: should we at least have something there until there is a consensus on the naming convention we should have? I'm trying to clear out an inconsistency where the wheels build job is named but the conda job isn't.

I can include the change in the PR description

ajschmidt8: that's fine

ajschmidt8 (approved):

PR #192, opened 2024-03-15
Drop CentOS 7 from workflows.

bdice:
This PR drops CentOS 7. See https://github.com/rapidsai/build-planning/issues/23.

PR #195, opened 2024-03-19
Remove input for build-2_28-wheels.

bdice:
Cleans up an extra option that was needed to support building both `manylinux_2_28` and `manylinux_2_17` wheels for cudf.

Depends on https://github.com/rapidsai/cudf/pull/15323.

### DIFF
--- a/.github/workflows/wheels-build.yaml
+++ b/.github/workflows/wheels-build.yaml
@@ -45,12 +45,6 @@ on:
         type: string
         default: ''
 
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Makes sense to me based on the linked issue and #192. I'm glad to see this workflow getting simplified.

vyasr (approved):
I guess at this point the var is already a no-op after #192 so it's just workflow cleanup that can happen once the variable is not in use.

PR #206, opened 2024-04-05
Disable fail-fast.

bdice:
This PR is an experiment to see if disabling `fail-fast` affects CI queue times. We plan to run this experiment for a week, and revert if CI queue times are noticeably worse.

We have seen a lot of test jobs being killed due to network errors, which triggers a `fail-fast` on other jobs. This wastes the resources of the jobs that were killed but didn't have network problems, because otherwise they could have run to completion and potentially passed. Frequently, if jobs are killed for "real" reasons like failed tests, the other jobs in that matrix are already scheduled and being run — and they are just about to hit the same failure, so the delta of time saved/lost by failing fast may not be significant.

The loss of information for devs when `fail-fast` is enabled is a cost to productivity. Waiting for jobs to queue and then only seeing partial information about failures is not ideal because it is difficult to see patterns like "this job always failed on ARM" or "this job always failed on CUDA 11."

### DIFF
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -98,7 +98,7 @@ jobs:
   tests:
     needs: compute-matrix
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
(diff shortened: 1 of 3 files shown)


pentschev (approved):
I'm +1 on that.

ajschmidt8 (approved):

PR #207, opened 2024-04-16
Add uploading artifact to `custom-job`

galipremsagar:
This change is necessary for the custom github status checks currently being run in `cudf` CI

### DIFF
--- a/.github/workflows/custom-job.yaml
+++ b/.github/workflows/custom-job.yaml
@@ -1,9 +1,5 @@
 on:
   workflow_call:
-    outputs:
-      job_output:
-        description: "job_output set by inputs.run_script"
-        value: "${{ jobs.build.outputs.job_output }}"
     inputs:
       build_type:
         required: true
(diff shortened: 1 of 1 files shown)


ajschmidt8 (commented):
How does this action behave if this file does not exist? Most jobs that use this shared-workflow probably won't generate this file.

ajschmidt8:
This section answered by question:

- https://github.com/actions/upload-artifact/?tab=readme-ov-file#customization-if-no-files-are-found

Here are the available options:

- https://github.com/actions/upload-artifact/blob/ef09cdac3e2d3e60d8ccadda691f4f1cec5035cb/action.yml#L11-L18

We should set it to `ignore` to minimize disruptions

galipremsagar:
> This section answered my question:
> 
> * https://github.com/actions/upload-artifact/?tab=readme-ov-file#customization-if-no-files-are-found
> 
> Here are the available options:
> 
> * https://github.com/actions/upload-artifact/blob/ef09cdac3e2d3e60d8ccadda691f4f1cec5035cb/action.yml#L11-L18
> 
> We should set it to `ignore` to minimize disruptions

Done 👍

ajschmidt8 (approved):

PR #210, opened 2024-05-09
Add `RAPIDS_AUX_SECRET_1` to select workflows

ajschmidt8:
There is a need to inject a secret into these shared-workflows by one of our repositories.

This PR adds an optional secret, `RAPIDS_AUX_SECRET_1`, to the necessary shared workflows.

I intentionally didn't add this secret to the other workflows, because its use is strongly discouraged. Therefore, we should only add it to workflows as needed.

This secret can be used by calling workflows like this:

```yaml
jobs:
  wheels-test:
    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@branch-24.06
    secrets:
      RAPIDS_AUX_SECRET_1: ${{ secrets.REPO_SPECIFIC_SECRET }}
```

### DIFF
--- a/.github/workflows/build-in-devcontainer.yaml
+++ b/.github/workflows/build-in-devcontainer.yaml
@@ -27,6 +27,11 @@ on:
         required: false
         type: string
         default: ''
+    # the use of secrets in shared-workflows is discouraged, especially for public repositories.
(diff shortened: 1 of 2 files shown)


ajschmidt8:
Ray approved this in Slack last week. admin-merging as a result of that approval.

PR #215, opened 2024-06-04
update codecov upload command

AyodeAwe:
This is an accompanying PR for the new `codecovcli` update [here](https://github.com/rapidsai/ci-imgs/pull/147).

Depends on https://github.com/rapidsai/ci-imgs/pull/147

### DIFF
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -147,10 +147,11 @@ jobs:
       - name: Run codecov
         if: inputs.run_codecov
         run: |
-          codecov \
-            -s \
-            "${RAPIDS_COVERAGE_DIR}" \
-            -v
+          codecovcli \
+            -v \
+            upload-process \
+            -C ${{ github.sha }} \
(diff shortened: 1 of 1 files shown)


.github/workflows/conda-python-tests.yaml
```diff
@@ -147,10 +147,12 @@ jobs:
...
-          codecov \
-            -s \
-            "${RAPIDS_COVERAGE_DIR}" \
-            -v
+          codecovcli \
+            -v \
+            upload-process \
+            -C ${{ github.sha }} \
+            -s "${RAPIDS_COVERAGE_DIR}" \
+            --handle-no-reports-found
```
AyodeAwe, line 155: This would apparently be necessary to match the behaviour in the previous uploader we used.

Example: this job. [Here](https://github.com/rapidsai/pynvjitlink/actions/runs/9358256535/job/25759882682#step:9:43), no coverage reports are found and the step should error, but the step passes. The implementation is apparently different in the new CLI tool, hence the need to add this `--handle-no-reports-found` flag.

raydouglass (approved):

PR #220, opened 2024-06-18
Remove verbosity added in #164

ajschmidt8:
This PR removes the verbosity that was added in #164. It was originally added for debugging purposes. It can be removed now. It causes our GitHub Actions logs to be really verbose.


jameslamb (approved):
I support this, I've noticed these too. Thanks!

PR #227, opened 2024-07-10
Always use rapidsai channel token.

bdice:
We would like to publish the conda packages for the `cugraph-ops` & `cumlprims_mg` repos to `rapidsai` instead of `nvidia` for releases to continue to reduce our dependence on the `nvidia` channel.


KyleFromNVIDIA (commented):
Are there any repos that are private that we still want to publish to `nvidia` instead of `rapidsai`? This change doesn't appear to discriminate.

bdice:
> Are there any repos that are private that we still want to publish to `nvidia` instead of `rapidsai`? This change doesn't appear to discriminate.

@raydouglass, please confirm. I believe we only want to publish packages to `rapidsai` now. To my knowledge there aren't any private repos that publish packages except for cumlprims_mg and cugraph-ops.

bdice:
I checked the list of `nvidia` packages here: https://anaconda.org/nvidia/repo

raydouglass (approved):

PR #228, opened 2024-07-11
Allow testing of earliest/latest dependencies.

bdice:
This PR allows the testing of earliest/latest dependencies. This is meant to be interpreted by each repository as it wishes, with CI scripts and `rapids-dependency-file-generator` to mediate which dependency versions are selected. For example, cuDF may wish to test early versions of `numpy` and `pandas`, while cuGraph may wish to test early versions of `networkx`, etc.

This is an experimental PR to prompt discussion / further work with @seberg.

### DIFF
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -105,9 +105,10 @@ jobs:
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
     runs-on: "linux-${{ matrix.ARCH }}-gpu-${{ matrix.GPU }}-${{ matrix.DRIVER }}-1"
     env:
-      RAPIDS_COVERAGE_DIR: ${{ github.workspace }}/coverage-results
-      RAPIDS_TESTS_DIR: ${{ github.workspace }}/test-results
       RAPIDS_ARTIFACTS_DIR: ${{ github.workspace }}/artifacts
+      RAPIDS_COVERAGE_DIR: ${{ github.workspace }}/coverage-results
+      RAPIDS_DEPENDENCIES: ${{ matrix.DEPENDENCIES }}
+      RAPIDS_TESTS_DIR: ${{ github.workspace }}/test-results
     container:
       image: rapidsai/ci-conda:cuda${{ matrix.CUDA_VER }}-${{ matrix.LINUX_VER }}-py${{ matrix.PY_VER }}
       options: ${{ inputs.container-options }}
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -71,22 +71,22 @@ jobs:
           export MATRICES="
             pull-request:
               # amd64
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'earliest', DEPENDENCIES: 'earliest' }
-              - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest'   }
+              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
+              - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               # arm64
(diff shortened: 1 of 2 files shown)


.github/workflows/wheels-test.yaml
```diff
@@ -108,15 +108,16 @@ jobs:
 
           echo "MATRIX=${MATRIX}" | tee --append "${GITHUB_OUTPUT}"
   test:
-    name: ${{ matrix.CUDA_VER }}, ${{ matrix.PY_VER }}, ${{ matrix.ARCH }}, ${{ matrix.LINUX_VER }}, ${{ matrix.gpu }}
+    name: ${{ matrix.CUDA_VER }}, ${{ matrix.PY_VER }}, ${{ matrix.ARCH }}, ${{ matrix.LINUX_VER }}, ${{ matrix.GPU }}, ${{ matrix.DRIVER }}-driver, ${{ matrix.DEPENDENCIES }}-deps
```
seberg, line 111: These name changes show up as:
screenshot_2024-07-16_at_16 34 22
(Maybe a bit long, but fits just about)

vyasr:
We should definitely proceed here because we need to do it sooner than we'll be able to [switch to uv](https://github.com/rapidsai/build-planning/issues/86), but JFYI in case it's useful for the future, [uv supports using the oldest instead of newest compatible versions via a flag](https://github.com/astral-sh/uv?tab=readme-ov-file#resolution-strategy). It won't be as surgical as what you are doing here though.

seberg:
Rebased.  Do we need to discuss if the pattern in the [RMM PR](https://github.com/rapidsai/rmm/pull/1613) pattern is nice enough to be pushed to most RAPIDS repos?

Using `uv` to automatically find the oldest versions sounds great, though.  Not sure the precision here has much up-sides in practice.  (Even if `uv` ends up `pip` only, testing this only in the wheels is maybe good enough in practice.).

vyasr:
TBH if we could use `uv` today, I might advocate just doing that because I agree that the precision may not have much upside other than restricting the complexity of your test space (it may, in fact, have downsides as a result). The problem is that we can't use uv quite yet as discussed in https://github.com/rapidsai/build-planning/issues/86 (due to https://github.com/rapidsai/build-planning/issues/85).

seberg:
I guess the question is then if we should wait for `uv` here if there is a clear ETA.  That would avoid back and forth on the `dependencies.yaml` (although I don't think it is terrible churn).

One thing wI could do is pull out the `conda` change, since I guess with `uv` we would probably limit this to just `pip` for simplicity.  (right now conda is nicer on the workflow side, but it seems tricky)

seberg:
Changes this to 24.10.  Should we go head with this, or redesign/wait for `uv`?

.github/workflows/conda-python-tests.yaml
```diff
@@ -66,27 +66,27 @@ jobs:
...
-              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest'   }
+              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
             nightly:
               # amd64
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'earliest' }
-              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'ubuntu20.04', GPU: 'v100', DRIVER: 'latest'   }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.0.1', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'latest'   }
-              - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest'   }
+              - { ARCH: 'amd64', PY_VER: '3.9',  CUDA_VER: '11.4.3', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
```
msarahan, line 80: Since the pattern for DRIVER is already 'earliest' and 'latest', can we keep that pattern for DEPENDENCIES also? In other words, just limit the values that a human has to remember?

msarahan (approved):
I think we should proceed now with this. It is not much churn if/when we use uv to achieve this purpose.

.github/workflows/conda-python-tests.yaml
seberg: Had changed it after the discussion [here](https://github.com/rapidsai/rmm/pull/1606#discussion_r1674557557), OTOH, also happy to change it again.  I agree here it is a bit of a dissonance.

jakirkham:
What else is still needed here?

jakirkham:
@msarahan what do you think?

bdice:
@seberg Resolve the PR conflicts and I think we can merge this.

jameslamb (approved):
I support this, thanks!

PR #229, opened 2024-07-11
Update CUDA version to 12.5.1

KyleFromNVIDIA:
Contributes to https://github.com/rapidsai/build-planning/issues/73

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -59,10 +59,10 @@ jobs:
           export MATRIX="
           # amd64
           - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04' }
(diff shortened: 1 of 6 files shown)


jameslamb (changes requested):
If we're going to drop ALL coverage of CUDA 12.2 in across building and testing, then we should also be replacing the CUDA 12.2 end-user container images with CUDA 12.5 images: https://github.com/rapidsai/build-planning/issues/73#issuecomment-2231443095

If we're going to keep publishing CUDA 12.2 container images for the 24.08 release, then I think we should preserve at least some coverage of 12.2.2 in nightly CI.

Have you already considered that? If not, then I think that's a conversation we need to have with all the build and ops folks.

I'll start a thread offline.

jameslamb (approved):
Thanks for the changes, and for keeping the matrices sorted 😁

raydouglass (approved):

AyodeAwe (approved):

jakirkham:
As we already switched all the projects back to `branch-24.08`, am going to delete `cuda-12.5.1`

If for some reason we missed something somewhere, now is the time to find out

Also if I goofed in some major way there is the restore button to bring it back

PR #230, opened 2024-07-15
Allow testing of earliest/latest dependencies.

seberg:
Mostly identical to gh-228, which allows the testing of oldest/latest dependencies.  What oldest/latest dependencies means exactly has to be set by each project in their `dependencies.yaml` file for the test env.

I am interested in it, because I would like to start testing with NumPy 2, but don't like the idea of stopping testing with 1.x at the same time.  But e.g. `cudf` testing with older pandas or cuGraph with older `networkx` would be good as well.

*Modifications from draft PR:*
- I renamed it to `oldest`, as suggested by Matthew.
- Noticed that the name is different between wheel and conda workflows, so modified both to include all matrix information.

(Marking as draft until I have linked PRs that show it working also for wheels, but Bradley did really all the work here, and it seems to work well.)

### DIFF
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -105,9 +105,10 @@ jobs:
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
     runs-on: "linux-${{ matrix.ARCH }}-gpu-${{ matrix.GPU }}-${{ matrix.DRIVER }}-1"
     env:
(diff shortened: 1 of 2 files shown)


seberg:
Hmmm, marking as non-draft either way.  I don't think I can use this workflow reasonably in other PRs unless the branch lives directly in the `shared workflows` repo.

bdice:
@seberg Correct — your fork won’t suffice for testing. I would recommend that you close this PR, request write access on this repo from ops, and push directly to my PR.

PR #233, opened 2024-07-25
Remove CUDA 12.2 conda builds.

bdice:
This PR removes CUDA 12.2 conda builds. There’s no need for those jobs because conda packages are only specific to the CUDA major version. If I remember the artifact naming scheme correctly, we run the risk of testing the wrong built artifacts if two minor versions from the same major family are used to build packages.

For testing, our intention is to build with 12.5 and use those packages for all CUDA 12 tests (12.0, 12.2, 12.5, etc.).

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -59,11 +59,9 @@ jobs:
           export MATRIX="
           # amd64
           - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04' }
-          - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04' }
           - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu22.04' }
           # arm64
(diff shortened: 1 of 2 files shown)


jakirkham:
Thanks Bradley! 🙏 

Would recommend reading this comment from James: https://github.com/rapidsai/shared-workflows/pull/229#pullrequestreview-2180936398

bdice:
We can retain testing coverage but not build coverage. We only need artifacts for one minor version due to CUDA Enhanced Compatibility. However, building multiple artifacts leads to clashes.

I am not on VPN at the moment to verify but I am pretty sure this is the case. Artifacts are named like `./linux-64/libcudf-24.08.00a390-cuda12_240725_g4aefcc7b29_390.tar.bz2` with no indication of the minor version. The choice between testing CUDA 12.2 builds and 12.5 builds is arbitrary, iirc, so we should provide only one build.

Keeping only one build per major CUDA version is the same way we handled 12.0 to 12.2 migration, and all past CUDA 11 migrations, iirc.

I’ll verify this further tomorrow.

jameslamb (approved):
You are 100% right! Here's what the artifact names look like for `cuml`, for example (only `cuda12`, no minor version):

image

https://downloads.rapids.ai/nightly/cuml/2024-07-25/6357c64/

Sorry, I'd asked that these build jobs be added in. Totally forgot about the artifact-naming thing and misunderstood the relationship between the versions we build against and the versions we test against.

KyleFromNVIDIA (approved):

jakirkham:
Thanks all! 🙏

Agree this is an important fix

PR #237, opened 2024-08-22
Check job results in pr-builder

KyleFromNVIDIA:
Add a new input, `needs`, which should just be the value of `${{ toJSON(needs) }}` from the calling workflow. This input will be checked to see if all of the needed jobs have a result of either `success` or `skipped`. This allows dependencies of `pr-builder` to be skipped. See https://github.com/rapidsai/cudf/pull/16642#discussion_r1727663594 for the motivating use case.

Workflows that use `pr-builder` should have `if: always()` and `with: needs: ${{ toJSON(needs) }}`. Workflows that have not yet been updated for this new input will continue to work as before, because the `needs` input has a default value of `{}`, which will always pass this check, just as if the old `exit 0` had been called.

### DIFF
--- a/.github/workflows/pr-builder.yaml
+++ b/.github/workflows/pr-builder.yaml
@@ -2,7 +2,7 @@ on:
   workflow_call:
     inputs:
       needs:
-        description: 'Value should always be `${{ needs }}`'
+        description: 'Value should always be `${{ toJSON(needs) }}`'
         required: false
         type: string
         default: '{}'
--- a/.github/workflows/pr-builder.yaml
+++ b/.github/workflows/pr-builder.yaml
@@ -2,7 +2,6 @@ on:
   workflow_call:
     inputs:
       needs:
-        description: "Value should always be `${{ '${{ toJSON(needs) }}' }}`"
         required: false
         type: string
         default: '{}'
--- a/.github/workflows/pr-builder.yaml
+++ b/.github/workflows/pr-builder.yaml
@@ -2,7 +2,7 @@ on:
   workflow_call:
     inputs:
       needs:
-        description: 'Value should always be `${{ toJSON(needs) }}`'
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-builder.yaml
```diff
@@ -1,9 +1,17 @@
...
 jobs:
   run:
     runs-on: ubuntu-latest
     steps:
       # This reusable workflow should depend on all other jobs in the calling workflow.
-      - run: exit 0
+      - name: "Check all dependent jobs"
+        env:
+          NEEDS: ${{ inputs.needs }}
+        run: echo "$NEEDS" | jq -e 'map(select((.result == "success" or .result == "skipped") | not)) | length == 0'
```
bdice, line 17: If I knew how, I'd write this as `any(result in {"success", "skipped"} for result in results)`. Specifically it'd be nice to short-circuit on the first non-match and use a set/list/array of acceptable statuses.

bdice (approved):
This appears to work well in the cudf PR that was testing it. Seems fine to me.

.github/workflows/pr-builder.yaml
KyleFromNVIDIA: Done

.github/workflows/pr-builder.yaml
```diff
@@ -1,9 +1,17 @@
...
 jobs:
   run:
     runs-on: ubuntu-latest
     steps:
       # This reusable workflow should depend on all other jobs in the calling workflow.
-      - run: exit 0
+      - name: "Check all dependent jobs"
+        env:
+          NEEDS: ${{ inputs.needs }}
+        run: echo "$NEEDS" | jq -e 'any((.result as $result | ["success", "skipped"] | any($result == .)) | not) | not'
```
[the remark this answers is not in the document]
KyleFromNVIDIA: https://github.com/rapidsai/shared-workflows/pull/238

PR #239, opened 2024-08-30
Add `changed-files` workflow

KyleFromNVIDIA:
The logic for the "figure out which files changed" workflow is rather complex. Establish a shared workflow to handle the complexity of it.

Contributes to https://github.com/rapidsai/build-planning/issues/94

Tested in https://github.com/rapidsai/cugraph/pull/4634 and https://github.com/rapidsai/cudf/pull/16713

### DIFF
--- a/.github/workflows/changed-files.yaml
+++ b/.github/workflows/changed-files.yaml
@@ -18,7 +18,7 @@ on:
           ) |
           from_entries
     outputs:
-      transformed_output:
+      changed_file_groups:
         value: ${{ jobs.changed-files.outputs.transformed_output }}
 
 defaults:
--- a/.github/workflows/changed-files.yaml
+++ b/.github/workflows/changed-files.yaml
@@ -1,0 +1,68 @@
+on:
+  workflow_call:
+    inputs:
+      files_yaml:
+        type: string
+        required: true
+      transform_expr:
+        type: string
+        required: true
+    outputs:
+      transformed_output:
+        value: ${{ jobs.changed-files.outputs.transformed_output }}
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  actions: read
+  checks: none
+  contents: read
+  deployments: none
+  discussions: none
+  id-token: write
+  issues: none
+  packages: read
+  pages: none
+  pull-requests: read
+  repository-projects: none
+  security-events: none
+  statuses: none
+
+jobs:
+  changed-files:
+    runs-on: ubuntu-latest
+    name: "Check changed files"
+    outputs:
+      transformed_output: ${{ steps.transform-changed-files.outputs.transformed-output }}
+    steps:
+      - name: Get PR info
+        id: get-pr-info
+        uses: rapidsai/shared-actions/get-pr-info@main
+      - name: Checkout code repo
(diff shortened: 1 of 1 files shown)


.github/workflows/changed-files.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Checkout code repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - name: Calculate merge base
+        id: calculate-merge-base
+        env:
+          PR_SHA: ${{ fromJSON(steps.get-pr-info.outputs.pr-info).head.sha }}
+          BASE_SHA: ${{ fromJSON(steps.get-pr-info.outputs.pr-info).base.sha }}
```
KyleFromNVIDIA, line 52: There is a possibility that this will result in this workflow reporting more files being changed than were actually changed. Per https://github.com/orgs/community/discussions/59677:

> The value of `${{ github.event.pull_request.base.sha }}` is not updated on every commit to the PR. It is the SHA of the base branch at the time the PR was created, which becomes out of date as the PR gets updated.

If the PR becomes out of date, and the author merges in the latest changes from the target branch, that could result in this workflow reporting the files changed in the target branch as changed in this PR.

.github/workflows/changed-files.yaml
```diff
@@ -0,0 +1,78 @@
...
+          from_entries
+    outputs:
+      changed_file_groups:
+        value: ${{ jobs.changed-files.outputs.transformed_output }}
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
```
bdice, line 28: Review this. Do we need “read” on packages?

bdice (approved):

.github/workflows/changed-files.yaml
KyleFromNVIDIA: Probably not. I'm also questioning whether we need `id-token: write`.

Requesting input from @ajschmidt8.

.github/workflows/changed-files.yaml
KyleFromNVIDIA: > But regardless, it's okay to be too aggressive and run more CI than strictly necessary.

That's a good point. I was simply thinking "the output will be incorrect" without regards to the harmless consequences of a false positive. We *don't* want false negatives, but I don't think this will result in that.

.github/workflows/changed-files.yaml
KyleFromNVIDIA: We can revisit this later. I'll go ahead and merge.

ajschmidt8: > Probably not. I'm also questioning whether we need `id-token: write`.
> 
> Requesting input from @ajschmidt8.

No, `id-token: write` is not necessary here. It's only needed in workflows that require AWS credentials via [aws-actions/configure-aws-credentials](https://github.com/aws-actions/configure-aws-credentials/).

Can this branch be deleted now that the PR is merged?

KyleFromNVIDIA: I went ahead and deleted it.

PR #247, opened 2024-09-27
add telemetry stuff for conda-cpp-build workflow

msarahan:
This is being tested in:

* cudf: https://github.com/rapidsai/cudf/pull/16924

When complete, this PR should add instrumentation to at least one workflow. More workflows can be adapted later once this is proven.

This is a draft PR until it is proven to be in working order. For cleanup, we'll also need to:

* [ ] remove the gha-tools stuff that is pulling from my personal fork, and merge https://github.com/rapidsai/gha-tools/pull/118 so the telemetry stuff is baked into ci-imgs.
* [ ] Move the OpenTelemetry install steps here into ci-imgs


msarahan:
I'm not sure I totally understand. Regardless of the presence of fluentbit, we can send span data to tempo. Fluentbit is the aggregator/forwarder, but we can point clients directly at the tempo instance. 

Right now, the main trace telemetry is simple retroactive parsing of the github job info. The only thing to turn off is really just not providing the endpoint. I designed things that the presence of the endpoint environment variable is what enabled telemetry. I think that was mostly in the rapids-otel-wrap script in gha-tools, though, and it's been a while since I played with that much, so it'd probably break. Nevertheless, the easiest way to turn off data collection is to just give it nowhere to go.

.github/workflows/conda-cpp-build.yaml
```diff
@@ -96,16 +121,60 @@ jobs:
...
+      # Temporary until https://github.com/rapidsai/ci-imgs/pull/196 is merged
+      - name: Install OpenTelemetry instrumentation
+        run: |
+          pip install opentelemetry-distro[otlp] \
+              opentelemetry-exporter-prometheus \
+              git+https://github.com/msarahan/opentelemetry-python-contrib.git@add-conda-build-instrumentation#subdirectory=instrumentation/opentelemetry-instrumentation-conda-build
+          opentelemetry-bootstrap -a install
+
+      - name: Setup proxy cache
+        uses: nv-gha-runners/setup-proxy-cache@main
```
ajschmidt8, line 146: was this addition intended?

I think we should save this for a separate PR to avoid introducing too many changes in this PR.

.github/workflows/wheels-publish.yaml
```diff
@@ -69,7 +69,7 @@ jobs:
         persist-credentials: false
 
     - name: Standardize repository information
-      uses: rapidsai/shared-actions/rapids-github-info@main
+      uses: ./shared-actions/rapids-github-info
```
ajschmidt8, line 72: I don't see a checkout for the `shared-actions` repository in this job.

ajschmidt8:
> I'm not sure I totally understand. Regardless of the presence of fluentbit, we can send span data to tempo. Fluentbit is the aggregator/forwarder, but we can point clients directly at the tempo instance.

Using a local Fluent Bit aggregator allows us to remove certificates from our workflows, which I want to do ASAP. I don't like having the certificates required in our workflows for security reasons. Security aside, the certs are short-lived and there is no good way to automatically rotate them as GH secrets. The Fluent Bit aggregator service has a sidecar container that handles certificate rotation automatically.

PR #248, opened 2024-10-02
Use Ubuntu 24.04 in CI workflows.

bdice:
This PR replaces `ubuntu22.04` with `ubuntu24.04`. This allows us to continue testing endpoints of our support matrix (we still support Ubuntu 20.04). The CI matrix has been reduced enough that I do not see a significant value proposition to keep any jobs with Ubuntu 22.04. It seems more valuable to have extensive testing of Ubuntu 24.04.

Depends on https://github.com/rapidsai/ci-imgs/pull/195.

Contributes to https://github.com/rapidsai/build-planning/issues/74.

### DIFF
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -69,18 +69,18 @@ jobs:
             pull-request:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
-              - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               # arm64
               - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
             nightly:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.4.3', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
-              - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               # arm64
-              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'oldest' }
-              - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
+              - { ARCH: 'arm64', PY_VER: '3.10', CUDA_VER: '12.2.2', LINUX_VER: 'ubuntu24.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'oldest' }
+              - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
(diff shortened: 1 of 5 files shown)


jameslamb (commented):
Thanks for getting this started. I left one suggestion for your consideration.

.github/workflows/wheels-test.yaml
```diff
@@ -80,18 +80,18 @@ jobs:
...
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest', DEPENDENCIES: 'oldest' }
+              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'v100', DRIVER: 'latest', DEPENDENCIES: 'oldest' }
               # arm64
               - { ARCH: 'arm64', PY_VER: '3.12', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest', DEPENDENCIES: 'latest' }
             nightly:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'latest', DEPENDENCIES: 'oldest' }
               - { ARCH: 'amd64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', GPU: 'v100', DRIVER: 'latest', DEPENDENCIES: 'latest' }
-              - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest', DEPENDENCIES: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'v100', DRIVER: 'latest', DEPENDENCIES: 'latest' }
```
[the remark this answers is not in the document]
bdice: This is a good argument. I agree that wheels are the right place to focus on due to more system dependency.

I looked at our CI usage dashboards and I think we could add this job to the "earliest" driver nodes without significant impact on our nightly CI queueing. The `linux-amd64-gpu-v100-latest-1` nightly queue is considerably longer, so putting this job on `linux-amd64-gpu-v100-earliest-1` would help alleviate some of that. I'll apply that for now, and I'll file some PRs to test these changes across RAPIDS 24.12.

```suggestion
              - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'earliest', DEPENDENCIES: 'latest' }
              - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'v100', DRIVER: 'latest', DEPENDENCIES: 'latest' }
```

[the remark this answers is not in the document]
jameslamb: Works for me! That's a good idea to the use the earliest-driver runner to spread this out more evenly, I like it.

bdice:
Some notes:
- I realized we can't use Ubuntu 24.04 in workflows with CUDA versions older than 12.5.1 because no nvidia/cuda images exist (we already excluded those in the ci-imgs matrix, I just hadn't considered the downstream effects).
  - I migrated the affected test jobs to Ubuntu 22.04.
- We should do our conda builds (C++ and Python) on Rocky Linux 8 because it's safer and better supported.
  - Rocky Linux 8 has the oldest glibc we support, which makes builds safer.
  - This will align with conda-forge as they migrate to glibc 2.28.
  - This also aligns RAPIDS conda builds with RAPIDS wheel builds, which already use Rocky Linux 8.
  - I think we have an old issue or thread about this somewhere -- we have meant to do switch conda builds to the oldest supported glibc in the past.

bdice:
CI testing is complete! Here is the link: https://github.com/rapidsai/build-planning/issues/74#issuecomment-2417959919

Are there any other repositories we should test? I think I got sufficient coverage here, and am happy to help iron out any additional issues that arise after merging.

bdice:
For future record, there were a couple of minor hiccups:
- cudf: missing timezone data needed by ORC sample data files, solved by installing `tzdata-legacy` into the CI image.
  - https://github.com/rapidsai/ci-imgs/pull/199
  - https://github.com/apache/orc/issues/2049
- raft/cuvs: issues with divide-by-zero in `awk`, solved by using a Python one-liner instead
  - https://github.com/rapidsai/cuvs/pull/414#issuecomment-2419830038
  - https://github.com/rapidsai/cuvs/pull/414/commits/94ff443de089e78ba061872ce65aae7c1b620b1b
  - https://github.com/rapidsai/raft/pull/2471/commits/d3b34c980385fb61b8ede2bfad399634dbf51321

jameslamb (approved):
Thanks! I'm happy with the matrix changes (they look identical to what I'd previously reviewed) and with the extent of the testing that was done.

I support merging this as-is, will defer to your judgment on the timing.

.github/workflows/conda-cpp-tests.yaml
```diff
@@ -69,18 +69,18 @@ jobs:
...
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
-              - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.5.1', LINUX_VER: 'ubuntu24.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
               # arm64
               - { ARCH: 'arm64', PY_VER: '3.11', CUDA_VER: '12.0.1', LINUX_VER: 'ubuntu20.04', GPU: 'a100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
             nightly:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.4.3', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
-              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu22.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
+              - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'ubuntu20.04', GPU: 'v100', DRIVER: 'latest',   DEPENDENCIES: 'latest' }
```
KyleFromNVIDIA, line 78: Why has this been downgraded from 22.04 to 20.04?

bdice: I rebalanced the OS version matrix a bit but this is open to change. As usual, our matrix should focus on the endpoints (oldest/newest). There is one job for Ubuntu 22.04 to provide additional coverage.

Would you prefer to make this 24.04?

bdice: Actually, I'm sorry. We cannot make this 24.04 because Ubuntu 24.04 only supports CUDA 12.5+.

Our options are 22.04 or 20.04. Rocky is already covered on the line above so we should keep this on Ubuntu.

bdice: For now, I'll revert to 22.04. I think that should be fine.

bdice: Ergh. I read this again on my terminal (somehow it escaped me when reading the GitHub diff).

If we change this to 22.04, we will have _only_ ARM coverage for 20.04. We should test this endpoint ("earliest supported Ubuntu") on amd64 as well.

KyleFromNVIDIA (approved):

PR #253, opened 2024-10-18
add package manager proxy cache for build jobs

msarahan:
Use new service, setup action at: https://github.com/nv-gha-runners/setup-proxy-cache

### DIFF
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -144,6 +144,8 @@ jobs:
       - name: Setup proxy cache
         uses: nv-gha-runners/setup-proxy-cache@main
         continue-on-error: true
+        # Skip the cache on RDS Lab nodes
+        if: ${{ matrix.GPU != 'v100' && matrix.GPU != 'a100' }}
 
       - name: Python tests
         run: ${{ inputs.script }}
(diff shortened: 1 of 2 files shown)


ajschmidt8:
Seeing these warnings in the downstream workflows.

image

Fortunately these arguments default to `true` anyway, so it should have no effect on this initial testing. But we'll want to fix it nonetheless

msarahan:
Rebased to 25.04 and removed the pip/conda specifiers that @ajschmidt8 pointed out.

ajschmidt8:
Can we get some new downstream tests to ensure everything is working as expected?

msarahan:
Looks like it is setting the env vars as expected. I don't know that I can comment on any performance improvements, but the work here at least looks non-disruptive.

bdice:
FYI, I merged the upstream so the PR diff would no longer show CUDA 12.8 related changes. Nothing functional should have changed.

ajschmidt8:
> Looks like it is setting the env vars as expected. I don't know that I can comment on any performance improvements, but the work here at least looks non-disruptive.
> 
> https://github.com/rapidsai/rmm/actions/runs/13057200836/job/36431299581?pr=1801

There's a Slack thread floating around where we identified that there may not be any particular performance enhancements since conda/PyPI could already be using CDNs. However, workflows _should_ be more reliable against network issues by using the proxy.

bdice (approved):

PR #257, opened 2024-10-31
Add breaking change Slack notifier

AyodeAwe:
Sends a notification to a designated Slack channel on every PR labelled with `breaking`, whenever any of the following events are triggered on the PR:
- `closed`
- `reopened`
- `labeled`
- `unlabeled`

This shared-workflow would be consumed in the following manner:

```
name: Trigger Breaking Change Notifications

on:
  pull_request:
    types:
      - closed
      - reopened
      - labeled
      - unlabeled

jobs:
  trigger-notifier:
    if: contains(toJSON(github.event.pull_request.labels.*.name), 'breaking')
    secrets: inherit
    uses: rapidsai/shared-workflows/.github/workflows/breaking-change-alert.yaml@<branch>
    with:
      sender_login: ${{ github.event.sender.login }}
      sender_avatar: ${{ github.event.sender.avatar_url }}
      repo: ${{ github.repository }}
      pr_number: ${{ github.event.pull_request.number }}
      pr_title: ${{ github.event.pull_request.title }}
      pr_body: ${{ github.event.pull_request.body || '_Empty PR description_' }}
      pr_base_ref: ${{ github.event.pull_request.base.ref }}
      pr_author: ${{ github.event.pull_request.user.login }}
      event_action: ${{ github.event.action }}
      pr_merged: ${{ github.event.pull_request.merged }}
```

**Example run**:
- Workflow run: https://github.com/rapidsai/literate-octo-potato/actions/runs/11615937445/job/32347710850?pr=854
- PR used: https://github.com/rapidsai/literate-octo-potato/pull/854


AyodeAwe:
After this PR is approved, I will open follow up PRs in all the relavant RAPIDS repositories.

KyleFromNVIDIA (commented):
~~I'm guessing the logic for calculating what constitutes a "breaking change" is not present in this workflow, but is left for the individual repos to decide? And this workflow just sends the Slack notification?~~

I see now that this workflow will be applied to PRs with the `breaking` label. I don't see a reference to a label event here, so I assume the logic of invoking the workflow on PRs with the label is left to the repos.

KyleFromNVIDIA (approved):

.github/workflows/breaking-change-alert.yaml
```diff
@@ -0,0 +1,122 @@
...
+    shell: bash
+
+jobs:
+  notify-breaking-changes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Determine notification parameters
+        id: notification
+        run: |
+          if [[ "${{ inputs.event_action }}" == "closed" && "${{ inputs.pr_merged }}" == "true" ]]; then
```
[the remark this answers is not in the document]
KyleFromNVIDIA: I thought about that too, but mostly for the second step below. At first I didn't think there was anything to be done there, but actually, now that I think about it, we should use this step to sanitize the inputs that go into the next step.

.github/workflows/breaking-change-alert.yaml
```diff
@@ -0,0 +1,122 @@
...
+      - name: Send Slack notification
+        uses: slackapi/slack-github-action@v1.27.0
+        with:
+          payload: |
+            {
+              "icon_emoji": "${{ steps.notification.outputs.icon }}",
+              "username": "RAPIDS Breaking Change Notifier - ${{ steps.notification.outputs.action }}",
+              "attachments": [
+                {
+                  "author_name": "${{ steps.notification.outputs.action }} PR by ${{ inputs.sender_login }}",
```
KyleFromNVIDIA, line 87: We should use the step above (or add a new step) to sanitize `inputs.*`, or at the very least escape them for JSON.

KyleFromNVIDIA (changes requested):

AyodeAwe:
> I don't see a reference to a label event here, so I assume the logic of invoking the workflow on PRs with the label is left to the repos.

Thanks for the review @KyleFromNVIDIA. 

Note this in the description of this PR:
```
name: Trigger Breaking Change Notifications

on:
  pull_request:
    types:
      - closed
      - reopened
      - labeled
      - unlabeled

# ...rest of the worfklow
```
Once this PR is approved and merged, I will open PRs across all RAPIDS repositories with that labelled event included, just as described in the PR description.

.github/workflows/breaking-change-alert.yaml
AyodeAwe: @KyleFromNVIDIA I presume that AJ's suggestion to move all inputs properties to intermediate environment variables should perhaps suffice for catering to injection attacks.

Do you think we should still sanitize after doing that?

.github/workflows/breaking-change-alert.yaml
```diff
@@ -0,0 +1,122 @@
...
+          elif [[ "${{ inputs.event_action }}" == "closed" ]]; then
+            echo "action=Closed" >> $GITHUB_OUTPUT
+            echo "color=#1d9bd1" >> $GITHUB_OUTPUT
+            echo "icon=:information_source:" >> $GITHUB_OUTPUT
+          elif [[ "${{ inputs.event_action }}" == "reopened" ]]; then
+            echo "action=Reopened" >> $GITHUB_OUTPUT
+            echo "color=warning" >> $GITHUB_OUTPUT
+            echo "icon=:warning:" >> $GITHUB_OUTPUT
+          else
+            action=${{ inputs.event_action }}
```
raydouglass, line 72: I think this needs to be changed/sanitized since it will be passed along to slack webhook.

Since it's the `else` case, maybe just something generic like `action=Changed`. Users can always go the PR for more details.

.github/workflows/breaking-change-alert.yaml
KyleFromNVIDIA: Environment variable substitution won't work in this step. This isn't a script, it's a JSON input that gets parsed by the external action. Someone could manipulate `inputs.sender_login` to make the external action do Bad Things. The way to mitigate that is to pass `inputs.sender_login` to a prior step, escape it as a JSON string, and then pass the escaped version to this, perhaps as `steps.sanitize.outputs.sender_login`.

AyodeAwe: I see what you're referring to now, thanks!

.github/workflows/breaking-change-alert.yaml
```diff
@@ -53,24 +53,51 @@ jobs:
   notify-breaking-changes:
     runs-on: ubuntu-latest
     steps:
+      - name: Escape inputs for JSON
+        id: escape
+        run: |
+          escape_json() {
+            echo "$1" | jq -R -c .
```
KyleFromNVIDIA, line 60: This adds quotes around everything that gets sanitized. We need to remove the quotes from the beginning and end.

AyodeAwe:
@KyleFromNVIDIA could I get a second review? The use of `awk` feels hacky but I think the function gets the job done. Thoughts?

.github/workflows/breaking-change-alert.yaml
```diff
@@ -0,0 +1,156 @@
...
+          SENDER_AVATAR: ${{ inputs.sender_avatar }}
+          REPO: ${{ inputs.repo }}
+          PR_NUMBER: ${{ inputs.pr_number }}
+          PR_TITLE: ${{ inputs.pr_title }}
+          PR_BODY: ${{ inputs.pr_body }}
+          PR_BASE_REF: ${{ inputs.pr_base_ref }}
+          PR_AUTHOR: ${{ inputs.pr_author }}
+        run: |
+          escape_json() {
+            echo "$1" | jq -R -c . | awk '{print substr($0, 2, length($0) - 2)}'
```
KyleFromNVIDIA, line 69: I would additionally add an assertion that there are actually quotes there:

```suggestion
            echo "$1" | jq -R -c . | grep -E '^".*"$' | awk '{print substr($0, 2, length($0) - 2)}' && [ "${PIPESTATUS[2]}" -eq 0 ]
```

AyodeAwe: Fixed

AyodeAwe:
@KyleFromNVIDIA any further thoughts?

KyleFromNVIDIA (approved):

raydouglass (approved):

PR #258, opened 2024-11-04
remove required flag from shared-actions param

msarahan:
This caused existing workflows that do not yet add this parameter to fail. Other workflows did not have this parameter as a required parameter.

These parameters are part of a rework of how shared actions get called. It is described in the readme in https://github.com/rapidsai/shared-actions

### DIFF
--- a/.github/workflows/conda-python-build.yaml
+++ b/.github/workflows/conda-python-build.yaml
@@ -36,13 +36,9 @@ on:
       shared_actions_repo:
         type: string
         description: git repo for rapidsai/shared-actions code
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Approving in case the answer to my question is "yes this is fine", as it seems that maybe you're trying to unblock CI in some other repo.

But don't we need to supply a default if there are repos out there not passing this configuration? What's going to happen if a null or empty string reaches this?

https://github.com/rapidsai/shared-workflows/blob/cb54c1e43e195f73438d8409f59526b95b87d318/.github/workflows/conda-python-build.yaml#L135-L140

PR #259, opened 2024-11-06
Escape multiline string

AyodeAwe:
Fixes a small issue with multiline strings passed to `pr_body` [here](https://github.com/rapidsai/shared-workflows/blob/branch-24.12/.github/workflows/breaking-change-alert.yaml#L26).

Follow up PR to #257

### DIFF
--- a/.github/workflows/breaking-change-alert.yaml
+++ b/.github/workflows/breaking-change-alert.yaml
@@ -66,7 +66,11 @@ jobs:
           PR_AUTHOR: ${{ inputs.pr_author }}
         run: |
           escape_json() {
-            echo "$1" | jq -R -c . | grep -E '^".*"$' | awk '{print substr($0, 2, length($0) - 2)}' && [ "${PIPESTATUS[2]}" -eq 0 ]
+            local flag="-R"
+            if [ "$2" = "stream" ]; then
+                flag="-sR"
+            fi
+            echo "$1" | jq ${flag} -c . | grep -E '^".*"$' | awk '{print substr($0, 2, length($0) - 2)}' && [ "${PIPESTATUS[2]}" -eq 0 ]
           }
 
           # Escape all input variables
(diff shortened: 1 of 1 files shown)


AyodeAwe:
cc @KyleFromNVIDIA 

Would love your review on this one since you reviewed #257

.github/workflows/breaking-change-alert.yaml
```diff
@@ -66,7 +66,11 @@ jobs:
           PR_AUTHOR: ${{ inputs.pr_author }}
         run: |
           escape_json() {
-            echo "$1" | jq -R -c . | grep -E '^".*"$' | awk '{print substr($0, 2, length($0) - 2)}' && [ "${PIPESTATUS[2]}" -eq 0 ]
+            local flag="-R"
+            if [ "$2" = "stream" ]; then
+                flag="-sR"
+            fi
+            echo "$1" | jq ${flag} -c . | grep -E '^".*"$' | awk '{print substr($0, 2, length($0) - 2)}' && [ "${PIPESTATUS[2]}" -eq 0 ]
```
KyleFromNVIDIA, line 73: I see no reason to differentiate between a single-line string and a multi-line string. The behavior is the same either way for a single-line string, and is an improvement for a multi-line string. Why not just treat it as multi-line unconditionally?

```suggestion
            echo "$1" | jq -sR -c . | grep -E '^".*"$' | awk '{print substr($0, 2, length($0) - 2)}' && [ "${PIPESTATUS[2]}" -eq 0 ]
```

.github/workflows/breaking-change-alert.yaml
```diff
@@ -75,7 +79,7 @@ jobs:
           echo "repo=$(escape_json "${REPO}")" >> $GITHUB_OUTPUT
           echo "pr_number=${PR_NUMBER}" >> $GITHUB_OUTPUT
           echo "pr_title=$(escape_json "${PR_TITLE}")" >> $GITHUB_OUTPUT
-          echo "pr_body=$(escape_json "${PR_BODY}")" >> $GITHUB_OUTPUT
+          echo "pr_body=$(escape_json "${PR_BODY}" stream)" >> $GITHUB_OUTPUT
```
KyleFromNVIDIA, line 82:
```suggestion
          echo "pr_body=$(escape_json "${PR_BODY}")" >> $GITHUB_OUTPUT
```

KyleFromNVIDIA (changes requested):

AyodeAwe:
Closing this PR and opening one from my fork instead.

.github/workflows/breaking-change-alert.yaml
AyodeAwe: @KyleFromNVIDIA left a comment [here](https://github.com/rapidsai/shared-workflows/pull/260#issuecomment-2460242280).

PR #260, opened 2024-11-06
Escape multiline string

AyodeAwe:
Fixes a small issue with multiline strings passed to `pr_body` [here](https://github.com/rapidsai/shared-workflows/blob/branch-24.12/.github/workflows/breaking-change-alert.yaml#L26).

Follow up PR to #257

### DIFF
--- a/.github/workflows/breaking-change-alert.yaml
+++ b/.github/workflows/breaking-change-alert.yaml
@@ -66,7 +66,11 @@ jobs:
           PR_AUTHOR: ${{ inputs.pr_author }}
         run: |
           escape_json() {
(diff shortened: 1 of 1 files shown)


KyleFromNVIDIA (changes requested):
Please see my review of #259.

AyodeAwe:
Thanks for the review @KyleFromNVIDIA 

The problem with not differentiating between a single-line string and a multi-line string is that for the single line strings, `\n` is appended to the output string. This is okay for a multiline string but not for the single line strings we expect to use in the json payload.

AyodeAwe:
See [here](https://github.com/rapidsai/literate-octo-potato/actions/runs/11706188305/job/32602709833#step:4:462) for an example.

KyleFromNVIDIA (approved):

PR #261, opened 2024-11-06
add if: always() to telemetry checkout steps

msarahan:
The telemetry additions here were masking other failed jobs because the telemetry was spread across too many steps. Only the last step had `if: always()`, when all of them should have it. We do want telemetry on failed jobs - that's why we have the `always()` set.


bdice (approved):
Changes look fine. Before merging, can you share how this was tested?

PR #262, opened 2024-11-08
refactor shared actions to simplify telemetry

msarahan:
Uses https://github.com/rapidsai/shared-actions/pull/21 to reduce the parameter passing. Instead of passing parameters as inputs, they are stored in a file that gets uploaded, then downloaded by later jobs.

PR #263, opened 2024-11-13
Revert "refactor shared actions to simplify telemetry"

msarahan:
Reverts rapidsai/shared-workflows#262

This does not handle repos that are not yet set up with the telemetry setup. Reverting until that can be fixed.


jameslamb (approved):

msarahan:
#264 is a better solution here. Closing. Test job at https://github.com/rapidsai/ucxx/actions/runs/11826502415/job/32952655898?pr=320

PR #264, opened 2024-11-13
add continue-on-error for telemetry-setup steps

msarahan:
Alternative to #263 - this is a placeholder to allow telemetry to fail gracefully while it is not rolled out to a given repo.


jameslamb (approved):
Yes this looks good to me. Noticed that the "Telemetry Summarize" steps at the end of these workflows already have `continue-on-error: true`, so hopefully with this PR all of the telemetry stuff will be non-blocking.

PR #271, opened 2025-01-23
Update to CUDA 12.8.0.

bdice:
Updates shared-workflows to use CUDA 12.8.0.

Depends on https://github.com/rapidsai/ci-imgs/pull/231.

### DIFF
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -69,7 +69,7 @@ jobs:
             pull-request:
               # amd64
               - { ARCH: 'amd64', PY_VER: '3.10', CUDA_VER: '11.8.0', LINUX_VER: 'rockylinux8', GPU: 'v100', DRIVER: 'earliest', DEPENDENCIES: 'oldest' }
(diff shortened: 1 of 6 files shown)


.github/workflows/conda-cpp-tests.yaml
vyasr, line 1: Given that we're keeping around one 12.2.2 job, should we also keep one 12.5.1 job as well?

vyasr: Same question applies to the other files.

vyasr (commented):
One question, but generally seems fine.

.github/workflows/conda-cpp-tests.yaml
vyasr: Oh I see that this is already being discussed in https://github.com/rapidsai/ci-imgs/pull/231#issuecomment-2613250628. Let's continue discussion over there instead of bifurcating here.

bdice: We don't have the resources to expand the number of jobs in the matrix any further. We can keep `[12.0, (12.2 or 12.5), 12.8]`. This looks like it will be `[12.0, 12.2, 12.8]` from the discussions on the other PR.

vyasr: I wasn't suggesting that we expand the matrix, just that we switch one of the 12.8s back to 12.5. In any case, I'll resolve this thread since the discussion is happening in ci-imgs.

vyasr (approved):

PR #273, opened 2025-01-28
Use NVKS runners

bdice:
This migrates amd64 CI jobs (PRs and nightlies) to use L4 GPUs from the NVKS cluster.

This is targeting `branch-25.04`.

A follow-up will be needed to add arm64 support once NVKS has ARM capacity.

### DIFF
--- a/.github/workflows/build-in-devcontainer.yaml
+++ b/.github/workflows/build-in-devcontainer.yaml
@@ -95,6 +95,10 @@ jobs:
         with:
           node-version: '16'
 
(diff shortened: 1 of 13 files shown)


jameslamb (approved):
oh nice! I've been keeping up with the conversations about this, agree with the plan.

bdice:
All repos have been migrated, except for the `docker` repo. However, that branch isn't ready for 25.04 yet (blocked by xgboost and `rapids=25.04` metapackages) so I am moving this forward.

PR #275, opened 2025-02-03
Allow `branch` option for `$BUILD_TYPE`

gforsyth:
Adds an additional option for the `*test.yaml` files for a specified `branch` option.

xref rapidsai/gha-tools#138


bdice (approved):
Yes, this is what I think we need! We'll have to iron out any issues once we test this with a repository.

PR #286, opened 2025-02-28
Use `twine` to upload wheels

KyleFromNVIDIA:
The `gh-action-pypi-publish` action is having problems with some repos. Try uploading with `twine` instead.

### DIFF
--- a/.github/workflows/wheels-publish.yaml
+++ b/.github/workflows/wheels-publish.yaml
@@ -47,7 +47,7 @@ permissions:
 jobs:
   wheel-publish:
     name: wheels publish
-    runs-on: linux-amd64-cpu4
+    runs-on: ubuntu-latest
     container:
       # CUDA toolkit version of the container is irrelevant in the publish step.
       # This just uploads already-built wheels to remote storage.
--- a/.github/workflows/wheels-publish.yaml
+++ b/.github/workflows/wheels-publish.yaml
@@ -100,9 +100,11 @@ jobs:
 
     - name: Publish the downloaded wheels to PyPI
       if: ${{ inputs.publish_to_pypi && steps.check_if_release.outputs.is_release_build == 'true' }}
-      uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        password: ${{ secrets.RAPIDSAI_PYPI_TOKEN }}
+      run: |
+        python3 -m pip install twine
+        python3 -m twine upload -u __token__ dist/*
+      env:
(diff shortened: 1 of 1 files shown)


jameslamb (approved):

jameslamb (commented):
> *The gh-action-pypi-publish action is having problems with some repos. Try uploading with twine instead.*

What are the problems, specifically? I'd like to know that before reviewing.

jameslamb:
Oh. I thought that specific problem was something we thought was related to running on self-hosted runners, and that we thought could be fixed by just switching to the GitHub Actions `ubuntu-latest` image.

I have one more idea we could try.... could we running `pypa/gh-action-pypi-publish` on `ubuntu-latest` **and not in a container**?

Removing this:

https://github.com/rapidsai/shared-workflows/blob/f1fb5d9e407d3307d6c4055387e7f6720a945fe7/.github/workflows/wheels-publish.yaml#L51-L56

And then doing whatever's required to make the relevant bits of `gha-tools` work outside of one of our CI images.

I just searched around and found some discussion suggesting that this type of issue might be something known and related to how that action sets paths when run in a container:

* https://github.com/pypa/gh-action-pypi-publish/pull/304
* https://github.com/pypa/gh-action-pypi-publish/issues/300

vyasr:
Yes, James is right, see https://github.com/rapidsai/pynvjitlink/pull/132#issuecomment-2707892727 most recently.

KyleFromNVIDIA:
Even if we did move it outside of the container, there are still other issues, like the fact that we're using it from a reusable workflow. I think we should use `twine` instead.

webknjaz:
> Even if we did move it outside of the container, there are still other issues, like the fact that we're using it from a reusable workflow. I think we should use `twine` instead.

You can do that at the cost of losing the ability to
1) do tokenless uploads (trusted publishing enables one to get rid of API tokens, so it's less secrets to maintain/scope/rotate)
2) improve the supply chain situation through standardized digital signatures (PEP 740)
3) display more bits of metadata as verified on the public PyPI project page

But even if you go this route, you can still improve security a little bit by storing the in-repo secrets within GitHub Environments scope as opposed to having them repo-global in the settings.

bdice (approved):
This worked for the pynvjitlink 0.5.2 release! Approving.

jakirkham:
Deleting `publish-wheels-on-github-runners` as it is no longer in use

(only `pynvjitlink` used and it has already switched back to `branch-25.04`)

Can always restore the branch if needed

PR #288, opened 2025-02-28
[Project Automation] Create High level tracker workflow + readme + helper script

jarmak-nv:
This pull request introduces a new GitHub Actions workflow for high-level project tracking and provides supporting documentation and scripts. The main changes include the addition of the workflow configuration, a detailed README explaining the workflow, and a helper script to obtain necessary project and field IDs.

### New GitHub Actions Workflow:
* [`.github/workflows/project-high-level-tracker.yaml`](diffhunk://#diff-8c67ae5cecfe18101499f48c1ea8dd35b2f6316a4ddad16035418c745baaad84R1-R268): Adds a comprehensive workflow to automate the management of GitHub Projects, including steps to process branch names, determine development stages, and update project fields based on release schedules.

### Documentation:
* [`docs/project-automation-readme.md`](diffhunk://#diff-dffa86e6fce5da83d1e273c3fbf7b58c11e3634d528afbc43c44445656160e3dR1-R163): Provides an overview of the new workflow, its architecture, and configuration details. It includes a mermaid diagram to illustrate the workflow and instructions on how to obtain necessary project and field IDs.

### Helper Script:
* [`docs/project-graphql-helper.py`](diffhunk://#diff-740571a71052b15b6f1609564e1c0d3e7c4212359c436c88ace4c337e1062b6cR1-R110): Adds a script to fetch project and field IDs from the GitHub GraphQL API, which are required for configuring the workflow.

### DIFF
--- a/.github/workflows/project-high-level-tracker.yaml
+++ b/.github/workflows/project-high-level-tracker.yaml
@@ -225,7 +225,7 @@ jobs:
 
   set-opened-date-field:
     uses: ./.github/workflows/project-set-text-date-numeric-field.yaml
-    if: github.event.action == 'opened'
+    if: github.event.action == 'opened' || github.event.action == 'workflow_dispatch'
     needs:
       - get-project-id
     with:
@@ -240,7 +240,7 @@ jobs:
 
   set-closed-date-field:
     uses: ./.github/workflows/project-set-text-date-numeric-field.yaml
-    if: github.event.action == 'closed'
+    if: github.event.action == 'closed' || github.event.action == 'workflow_dispatch'
     needs:
       - get-project-id
     with:
--- a/.github/workflows/project-high-level-tracker.yaml
+++ b/.github/workflows/project-high-level-tracker.yaml
@@ -32,23 +32,23 @@ on:
 
       # ------------- Option IDs -------------
       # Stage options
-      STAGE_PLANNED_ID:
+      STAGE_PLANNED_NAME:
         description: "📓 Planned option ID"
         default: "📓 Planned"
         type: string
-      STAGE_DEVELOPMENT_ID:
+      STAGE_DEVELOPMENT_NAME:
         description: "💻 Development option ID"
         default: "💻 Development"
         type: string
-      STAGE_BURNDOWN_ID:
+      STAGE_BURNDOWN_NAME:
         description: "🔥 Burndown option ID"
         default: "🔥 Burndown"
         type: string
-      STAGE_CODE_FREEZE_ID:
+      STAGE_CODE_FREEZE_NAME:
         description: "❄ Code Freeze option ID"
         default: "❄ Code Freeze"
         type: string
-      STAGE_HOTFIX_ID:
+      STAGE_HOTFIX_NAME:
         description: "🩹 Hotfix option ID"
         default: "🩹 Hotfix"
         type: string
@@ -160,22 +160,22 @@ jobs:
           echo "Release date: $RELEASE_DATE"
           
           # Determine stage based on current date
-          STAGE_ID="${{ inputs.STAGE_PLANNED_ID }}"
+          STAGE_ID="${{ inputs.STAGE_PLANNED_NAME }}"
           
           if [[ "$CURRENT_DATE" < "$DEV_START" ]]; then
-            STAGE_ID="${{ inputs.STAGE_PLANNED_ID }}"
+            STAGE_ID="${{ inputs.STAGE_PLANNED_NAME }}"
             echo "Stage: Planned"
           elif [[ "$CURRENT_DATE" < "$BURNDOWN_START" ]]; then
-            STAGE_ID="${{ inputs.STAGE_DEVELOPMENT_ID }}"
+            STAGE_ID="${{ inputs.STAGE_DEVELOPMENT_NAME }}"
             echo "Stage: Development"
           elif [[ "$CURRENT_DATE" < "$FREEZE_START" ]]; then
-            STAGE_ID="${{ inputs.STAGE_BURNDOWN_ID }}"
+            STAGE_ID="${{ inputs.STAGE_BURNDOWN_NAME }}"
             echo "Stage: Burndown"
           elif [[ "$CURRENT_DATE" < "$RELEASE_DATE" ]]; then
-            STAGE_ID="${{ inputs.STAGE_CODE_FREEZE_ID }}"
+            STAGE_ID="${{ inputs.STAGE_CODE_FREEZE_NAME }}"
             echo "Stage: Code Freeze"
           else
-            STAGE_ID="${{ inputs.STAGE_HOTFIX_ID }}"
+            STAGE_ID="${{ inputs.STAGE_HOTFIX_NAME }}"
             echo "Stage: Hotfix"
           fi
           
--- a/.github/workflows/project-high-level-tracker.yaml
+++ b/.github/workflows/project-high-level-tracker.yaml
@@ -1,0 +1,268 @@
+name: RAPIDS High Level Tracking
+
+on:
+  workflow_call:
+    inputs:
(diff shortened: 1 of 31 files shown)


.github/workflows/project-high-level-tracker.yaml
```diff
@@ -0,0 +1,268 @@
+name: RAPIDS High Level Tracking
+
+on:
+  workflow_call:
+    inputs:
+      PROJECT_ID:
+        description: "Project ID"
+        default: "PVT_kwDOAp2shc4ArK9w"
```
jameslamb, line 8: General comment for all of these `default:` values with what look like identifiers from a service.... what are they? Where did the values come from? Are they intended to be overridden?

If every user of this is expected to provide these values at the site where this is called, then I think that:

* the defaults should be removed
* the inputs should all have `required: true`
* docs should be added here explaining how to set values for them (either how to find the things to hard-code, or how to get these via templating with the various runtime contexts GitHub Actions provides: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context)

If these values are intended to be used literally and never overridden, then I think that:

* they should be moved out of inputs and instead either hard-coded below or shared via the `env.` context (https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#env-context)

If these values *could* be used literally but *might* be overridden, then I think that:

* the inputs should all have `required: true`
* docs should be added here explaining how to set values for them
* code comments should be added explaining where these defaults came from

.github/workflows/project-high-level-tracker.yaml
```diff
@@ -0,0 +1,268 @@
...
+        default: "PVTF_lADOAp2shc4ArK9wzgiTHmQ"
+        type: string
+      END_DATE_FIELD_ID:
+        description: "End Date Field ID"
+        default: "PVTF_lADOAp2shc4ArK9wzgiTHmU"
+        type: string
+
+      # ------------- Option IDs -------------
+      # Stage options
+      STAGE_PLANNED_ID:
```
jameslamb, line 35: I'm a little confused by the naming with `*_ID` for these things... they look like names or labels. I think changing the input names to something like `*_NAME` would make it clearer that these are free-text display names that don't have uniqueness constraints.

Am I right about that? If so, would you consider that naming change?

docs/project-automation-readme.md
```diff
@@ -0,0 +1,164 @@
...
+The automations can also synchronize fields from PR to Issue such as the statuses, release information, and development stages into project fields, making project management effortless and consistent.
+
+## 🔍 The Challenge
+
+GitHub Projects are powerful, but automating them is not straightforward:
+
+- Each item has a **project-specific ID** that's different from its global node ID
+- Different field types (text, date, single-select, iteration) require different GraphQL mutations
+- Linking PRs to issues and keeping them in sync requires complex operations
+   - Work often begins in an Issue, but then is completed in the PR, requiring this synchronization
```
jameslamb, line 16: I don't understand these workflows and the GraphQL code very well, so apologies if this is an ignorant question... does any of this automation **assume** that 1 issue will be resolved by exactly 1 PR?

The language here (talking about "the" PR) makes me think that it might.

I think that's an assumption that's likely to be violated often, especially for large efforts like https://github.com/rapidsai/cuvs/issues/735

jameslamb (commented):
I took a first pass at this, will do a more thorough review next week. Just publishing a non-blocking review now so you can see my first few comments.

.github/workflows/project-high-level-tracker.yaml
jarmak-nv: Yeah we can probably change these; we use the field name as the ID in the query (since they added the capability to do so semi-recently).

Making a note here to change it so it's clearer.

docs/project-automation-readme.md
jarmak-nv: Yeah totally agree it's a little confusing; for any given workflow-triggering PR the `update-linked-issues` workflow synchronizes all issues that are linked to that PR.

Ultimately, GitHub itself doesn't have a strong mechanism for delineating the 1-to-many relationship of PRs and Issues in their linkages. If you link 5 PRs to 1 issue, closing _any_ of the PRs closes _all_ of the issues.

The linkage between the two lives inside the PR, so I say "the PR" referring to the PR that triggers the workflow, but can totally see changing it or adding some clarity!

.github/workflows/project-high-level-tracker.yaml
jarmak-nv: Updated in https://github.com/rapidsai/shared-workflows/pull/288/commits/1d7d605961b1d5a0a418752e08953f935d1fb28f

docs/project-automation-readme.md
jarmak-nv: Added some text, lmk if it adds any clarity

.github/workflows/project-high-level-tracker.yaml
```diff
@@ -0,0 +1,268 @@
...
+            echo "Stage: Development"
+          elif [[ "$CURRENT_DATE" < "$FREEZE_START" ]]; then
+            STAGE_ID="${{ inputs.STAGE_BURNDOWN_NAME }}"
+            echo "Stage: Burndown"
+          elif [[ "$CURRENT_DATE" < "$RELEASE_DATE" ]]; then
+            STAGE_ID="${{ inputs.STAGE_CODE_FREEZE_NAME }}"
+            echo "Stage: Code Freeze"
+          else
+            STAGE_ID="${{ inputs.STAGE_HOTFIX_NAME }}"
+            echo "Stage: Hotfix"
```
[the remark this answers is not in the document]
jarmak-nv: Yeah fair, if these stay as variables vs going hard-coded per James' comment above I'll update.

I don't think this is a vector for script injection though because the only way these variables get set is inside the action (they aren't affected by a user defined field in a workflow-triggering PR). If a bad actor attempted to set the values in their own branch, it wouldn't have an effect unless we merged it.

One of the frustrating parts of these workflows is that testing is challenging because they need to be merged into the default branch to be functional, but it does provide protection against script injection as a benefit in many cases!

.github/workflows/project-high-level-tracker.yaml
jameslamb: > *anyone can query those values.*

Ok but *how*? Let's imagine you're not available and this breaks complaining about a project with this ID not existing... how would I go find the appropriate values?

The other thing I'm confused about... if it's ok to hard-code a single project's ID, then why is this workflow getting checked into `shared-workflows` (which contains workflows that are intended to be re-used in many places), instead of in whatever repo the project board is associated with?

docs/project-automation-readme.md
jameslamb: > *Added some text, lmk if it adds any clarity*

> *If you link 5 PRs to 1 issue, closing any of the PRs closes all of the issues.*

Thanks. I re-read the docs today and it's good to see they no longer imply a 1-to-1 relationship between PRs and issues.

I think the biggest thing I'm still really unsure about it what you mean (in comments here and in the docs) by "link to".

For example, if I leave a comment like `"Contributes to #17963"` in a PR's description or in a comment, I have "linked" the PR to the issue.. GitHub will generate an update on the corresponding issues like these:

image

*(ref: https://github.com/microsoft/LightGBM/issues/3867)*

But those types of linkages don't result in the behavior "closing the PR closes the issue".

I suspect you mean specifically using one of these keywords: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue

If so, I'd add a link to that in the docs.

jarmak-nv:
Closing this PR;  docs are still useful and have been updated/moved to #415

Issue #302, opened 2025-03-17
Remove use of tj-actions/changed-files


bdice:
Due to a security vulnerability ([CVE-2025-30066](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2025-30066)) which took down `tj-actions/changed-files`, we must disable the use of that action. In the meantime, pull request CI is blocked across RAPIDS because it relies on this action.

I am going to try to institute a temporary fix that will always say that files were changed, and thus always run CI. A longer-term fix will be to replace this action with a suitable alternative so that we can regain the benefits of skipping unnecessary CI jobs.

cc: @KyleFromNVIDIA (author of changed files logic)

KyleFromNVIDIA:
I'm working on a fix now. My fix will report an error if they specify `transform_expr` and direct them to this issue.

jameslamb:
@KyleFromNVIDIA did this in #309

There are no remaining uses of `tj-actions/changed-files` in this repo or in any other active repo in RAPIDS. I've added this to the build-infra board and put it in the 25.04 release for historical context.

PR #304, opened 2025-03-17
Report all files as changed temporarily

KyleFromNVIDIA:
Issue: https://github.com/rapidsai/shared-workflows/issues/302

### DIFF
--- a/.github/workflows/changed-files.yaml
+++ b/.github/workflows/changed-files.yaml
@@ -51,41 +51,26 @@ jobs:
         uses: rapidsai/shared-actions/telemetry-dispatch-setup@main
         continue-on-error: true
         if: ${{ vars.TELEMETRY_ENABLED == 'true' && github.run_attempt == '1' }}
-      - name: Get PR info
-        id: get-pr-info
-        uses: nv-gha-runners/get-pr-info@main
-      - name: Checkout code repo
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-          persist-credentials: false
-      - name: Calculate merge base
-        id: calculate-merge-base
-        env:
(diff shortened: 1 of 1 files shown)


ajschmidt8 (changes requested):
We should remove `id-token: write` from this workflow since it's not needed.

.github/workflows/changed-files.yaml
```diff
@@ -51,41 +51,28 @@ jobs:
...
+            select(.key | endswith("_any_changed")) |
+            {
+              "key": (.key | rtrimstr("_any_changed")),
+              "value": .value,
+            }
+          ) |
+          from_entries'
+          echo "Default transform expression: $default_transform_expr"
+          echo "Our transform expression: $TRANSFORM_EXPR"
+          if [ "$TRANSFORM_EXPR" != "$default_transform_expr" ]; then
```
bdice, line 71: Perhaps we can simplify and set the default transform expression to `"default"` and fail if it's not `"default"`. That way we don't have to compare multi-line strings.

I would also accept a hard break that removes that parameter from the workflow, because we do not have any uses of that parameter that I could find in the `rapidsai` org.

KyleFromNVIDIA: I ran into string comparison issues, so I just removed it from the workflow.

KyleFromNVIDIA:
CI is passing at https://github.com/rapidsai/cudf/pull/18294. I think this is ready.

bdice (approved):

rgsl888prabhu:
@KyleFromNVIDIA Is there a plan to push this change to 25.02 ?

jakirkham:
Done in PR: https://github.com/rapidsai/shared-workflows/pull/314

PR #276, opened 2025-02-04
Use nightly matrix for branch tests

bdice:
This PR (built on the `nvks-runners` branch) uses the nightly matrix for branch tests.

PR #308, opened 2025-03-18
make matrix type configurable

jameslamb:
Closes https://github.com/rapidsai/build-planning/issues/158

Today, the matrix (combinations of job variants) for testing workflows is determined by the "`build_type`" (one of "branch", "nightly", or "pull-request").

This proposes:

* making the matrix selection configurable
  - *so you could for example achieve behavior like "run the nightly test matrix in PR CI"*
* adding descriptions on the enum-like inputs, listing the valid values
* expanding input validation
  - *to ensure unexpected values result in an immediate, clear, loud error*

### How I tested this

See the comments on this PR I used to test: https://github.com/rapidsai/rmm/pull/1865

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -2,6 +2,7 @@ on:
   workflow_call:
     inputs:
       build_type:
+        description: "One of: [branch, nightly, pull-request]"
         required: true
         type: string
       branch:
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -3,11 +3,7 @@ on:
     inputs:
       build_type:
         required: true
-        type: choice
-        options:
-          - branch
-          - nightly
-          - pull-request
+        type: string
       branch:
         type: string
       date:
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -3,7 +3,11 @@ on:
     inputs:
       build_type:
         required: true
-        type: string
+        type: choice
+        options:
+          - branch
+          - nightly
+          - pull-request
       branch:
         type: string
       date:
--- a/.github/workflows/conda-cpp-tests.yaml
+++ b/.github/workflows/conda-cpp-tests.yaml
@@ -2,21 +2,14 @@ on:
   workflow_call:
     inputs:
       build_type:
+        description: "One of: [branch, nightly, pull-request]"
         required: true
-        type: choice
-        options:
-          - branch
-          - nightly
-          - pull-request
+        type: string
       matrix_type:
-        required: true
-        type: choice
-        options:
-          # "auto" = choose a value based on what's provided via 'build_type'
-          - auto
-          - nightly
-          - pull-request
-        default: auto
+        description: "One of: [auto, nightly, pull-request]. 'auto' means 'choose a value based on what's provided via build_type'."
+        required: false
+        type: string
+        default: "auto"
       branch:
         type: string
       date:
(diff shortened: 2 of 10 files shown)


.github/workflows/conda-cpp-build.yaml
```diff
@@ -3,7 +3,11 @@ on:
     inputs:
       build_type:
         required: true
-        type: string
+        type: choice
+        options:
+          - branch
+          - nightly
+          - pull-request
```
jameslamb, line 10: Sad, looks like `choice` inputs can't be used with the `workflow_call` trigger

I'd hoped we could use GitHub Actions `choice` type to implement enum-like behavior ([GitHub docs](https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/workflow-syntax-for-github-actions#example-of-onworkflow_dispatchinputs)), so that `matrix_type` and other inputs didn't have to be arbitrary strings.

Looks like that's not possible 😭 

> The workflow is not valid. In .github/workflows/pr.yaml (Line: 89, Col: 11): Error from called workflow rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@customize-matrix (Line: 6, Col: 15): Unexpected value 'choice' In .github/workflows/pr.yaml (Line: 89, Col: 11): Error from called workflow rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@customize-matrix (Line: 7, Col: 9): Unexpected value 'options'

([run link from RMM test PR](https://github.com/rapidsai/rmm/actions/runs/13926298643))

To be fair, `actionlint` also reported that but I'd hoped (naively) that it was just wrong or maybe confused because these are shared workflows. Sigh 😭

.github/workflows/conda-cpp-tests.yaml
```diff
@@ -2,8 +2,14 @@ on:
   workflow_call:
     inputs:
       build_type:
+        description: "One of: [branch, nightly, pull-request]"
```
ajschmidt8, line 5: This and other test workflows only get invoked after a pull-request build or nightly build, not after a branch build.

This decision is deliberate to avoid straining our GPU capacity.

Because of this, I'm not sure including a `branch` build type here makes sense since it will never get used.

.github/workflows/conda-cpp-tests.yaml
```diff
@@ -85,10 +96,13 @@ jobs:
...
           "
 
-          # Use the nightly matrix for branch tests
-          MATRIX_TYPE="${BUILD_TYPE}"
-          if [[ "${MATRIX_TYPE}" == "branch" ]]; then
-            MATRIX_TYPE="nightly"
+          # only overwrite MATRIX_TYPE if it was set to 'auto'
+          if [[ "${MATRIX_TYPE}" == "auto" ]]; then
+            MATRIX_TYPE="${BUILD_TYPE}"
+            # Use the nightly matrix for branch tests
```
ajschmidt8, line 102: As mentioned above, we currently don't run tests after branch builds.

ajschmidt8 (commented):
I left a comment on `.github/workflows/conda-cpp-tests.yaml`, but it's relevant to all the test workflow changes.

.github/workflows/conda-cpp-tests.yaml
vyasr: Making this possible is actually one of the goals here. See https://github.com/rapidsai/gha-tools/issues/138.

jameslamb: To be clear, I'm not proposing something **new** here.

Currently on `branch-25.04`, there's already handling of `build_type: branch` in this workflow and the others touched by this PR.

https://github.com/rapidsai/shared-workflows/blob/9ebdde271fab64e3752b58b5b9878d804f69b6af/.github/workflows/conda-cpp-tests.yaml#L88-L93

This change just documents the current state.

I'm happy to try to make this stricter as part of this PR if you think there aren't any uses out there that trigger tests for `branch` builds... given what I just wrote, do you want me to do that?

jameslamb: Wait actually, looking through the blame... the code I linked was just added around a month ago, by @bdice : #276

I think to support manually triggering a `test` job via `workflow_dispatch`.

ajschmidt8: Whoops. I see that now. I misinterpreted the diff on your PR.

I think my comment can be disregarded than.

ajschmidt8 (approved):

.github/workflows/conda-cpp-tests.yaml
```diff
@@ -87,10 +98,13 @@ jobs:
...
+            if [[ "${BUILD_TYPE}" == "branch" ]]; then
+              MATRIX_TYPE="nightly"
+            fi
```
vyasr, line 107: Very minor nit: This feels like it should be an if-else.

vyasr (approved):
Logic looks fine to me.

PR #311, opened 2025-03-18
Add workflow to update `latest` and `default` branches

KyleFromNVIDIA:
Tested in https://github.com/rapidsai/cudf/pull/18308.

### DIFF
--- a/.github/workflows/update-latest-branch.yaml
+++ b/.github/workflows/update-latest-branch.yaml
@@ -32,9 +32,9 @@ jobs:
             const tryUpdateBranch = async (to, from) => {
               if (currentBranch != to) {
                 if (dryRun) {
-                  console.log(`Would have skipped branch "${to}" since it is not the current branch ("${currentBranch}"")`);
+                  console.log(`Would have skipped branch "${to}" since it is not the current branch ("${currentBranch}")`);
                 } else {
-                  console.log(`Skipping branch "${to}" since it is not the current branch ("${currentBranch}"")`);
+                  console.log(`Skipping branch "${to}" since it is not the current branch ("${currentBranch}")`);
                 }
                 return;
               }
--- a/.github/workflows/update-latest-branch.yaml
+++ b/.github/workflows/update-latest-branch.yaml
@@ -3,7 +3,7 @@ name: Update `latest` branch
 on:
   workflow_call:
     inputs:
-      current_ref:
+      branch:
         type: string
         required: true
 
@@ -15,11 +15,10 @@ jobs:
         id: update-latest-branch
         uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
         env:
-          CURRENT_REF: ${{ inputs.current_ref }}
+          BRANCH: ${{ inputs.branch }}
         with:
           script: |
-            const { CURRENT_REF: currentRef } = process.env;
-            const currentBranch = /^refs\/heads\/(.*)$/.exec(currentRef)[1];
+            const { BRANCH: currentBranch } = process.env;
 
             // Get all branches
             const branches = await github.paginate(github.rest.repos.listBranches, context.repo);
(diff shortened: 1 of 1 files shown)


.github/workflows/update-latest-branch.yaml
```diff
@@ -0,0 +1,91 @@
...
+                  console.log(`Skipping branch "${to}" since it is not the current branch ("${currentBranch}")`);
+                }
+                return;
+              }
+
+              const ref = await github.rest.git.getRef({ ...context.repo, ref: `heads/${to}` });
+              const sha = ref.data.object.sha;
+
+              if (branches.some((b) => b.name == from)) {
+                if (dryRun) {
```
bdice, line 46: Maybe it would be cleaner to do something like:

```js
// At the top, announce that this is a dry run
if (dryRun) {
  console.log(`This is a dry run. No actions will be taken.`);
}

// For each action, output the logs unconditionally but only act if `!dryRun`
console.log(`Updating branch...`);
if (!dryRun) {
  await github.rest.git.updateRef(...);
}
```

KyleFromNVIDIA: Done: https://github.com/rapidsai/cudf/actions/runs/13952694344/job/39057321557?pr=18308

.github/workflows/update-latest-branch.yaml
```diff
@@ -0,0 +1,91 @@
...
+              }
+              if (next.major < prev.major) {
+                return prev;
+              }
+              if (next.minor > prev.minor) {
+                return next;
+              }
+              return prev;
+            }, null).name;
+            await tryUpdateBranch(latestBranch, "latest");
```
bdice, line 82: With the current structure, `tryUpdateBranch` needs three pieces of information: the branch that triggered this action (current branch), the "to" branch (e.g. `branch-25.06`), and the "from" branch (`latest`).

The implementation is basically: `if ${currentBranch} == ${to}, then update the head of ${from} to match the head of ${to}`. Is that right?

The naming and logic feel backwards to me, if I am reading that correctly.

Maybe we can choose a different set of names, like `if ${currentBranch} == ${source}, then update the head of ${target} to match the head of ${source}`. Maybe I'm just reading it funny.

I do think we could pull out `${currentBranch}` so that it isn't needed inside of `tryUpdateBranch` at all. That should only need two pieces of information (to/from) and the check for `if ${currentBranch} == ${to}` should happen outside of the update function. Maybe something like this:

```js
// Do the update. No conditions.
const updateBranch = async (to, from) => { ... }

// Check whether to update, and maybe update.
const tryUpdateBranch = async(current, to, from) => { ... }

// Wrap logic for getting the latest branch in a function
const getLatestBranch = (branches) => { ... }
```

.github/workflows/update-latest-branch.yaml
```diff
@@ -0,0 +1,91 @@
...
+                  console.log(`Would have updated branch "${from}" to point to branch "${to}", SHA ${sha}`);
+                } else {
+                  console.log(`Updating branch "${from}" to point to branch "${to}", SHA ${sha}`);
+                  await github.rest.git.updateRef({ ...context.repo, ref: `heads/${from}`, sha, force: true });
+                }
+              } else {
+                if (dryRun) {
+                  console.log(`Would have created branch "${from}" to point to branch "${to}", SHA ${sha}`);
+                } else {
+                  console.log(`Creating branch "${from}" to point to branch "${to}", SHA ${sha}`);
```
bdice: I'm not sure if I want to let this bot create branches. If the bot creates a branch, it won't have the right protections. Perhaps we should require a manual step from ops to create the branch first, and set the right protections.

KyleFromNVIDIA:
@bdice I've taken your feedback, removed an unnecessary API call, and added an `all` input to update all branches instead of just the one that got pushed. You can see the results [here](https://github.com/rapidsai/cudf/actions/runs/13953603941/job/39059181731?pr=18308).

bdice (approved):
Love it! This is really nice.

raydouglass (approved):

bdice:
@KyleFromNVIDIA Following up on this -- is there an issue tracking how this will be rolled out / used in RAPIDS? I don't see one cross-linked here.

KyleFromNVIDIA:
Oh, but we also realized that the token in the workflow won't have permission to push to protected branches, and we need an app or something.

PR #318, opened 2025-03-19
make `RAPIDS_AUX_SECRET_1` be an input variable secret name, not the secret itself

trxcllnt:
Switches the `RAPIDS_AUX_SECRET_1` value to an input of a secret name, then defines the `RAPIDS_AUX_SECRET_1` envvar from the `secrets` context to align with the way we're passing in other secrets.

Without this change, it isn't possible for a calling workflow to define both `inputs.extra-repo-deploy-key` and `secrets.RAPIDS_AUX_SECRET_1` -- the caller has to choose either:
1. `{ secrets: inherit, with: { extra-repo-deploy-key: 'KEY_NAME' } }`, or
2. `{ secrets: { RAPIDS_AUX_SECRET_1: ${{ secrets.SECRET }} } }`

In the second case, the shared workflow's `secrets` context only contains `RAPIDS_AUX_SECRET_1`, not `KEY_NAME`, because we had to pass it instead of `secrets: inherit`.

### DIFF
--- a/.github/workflows/build-in-devcontainer.yaml
+++ b/.github/workflows/build-in-devcontainer.yaml
@@ -30,11 +30,11 @@ on:
         required: false
         type: string
         default: ''
-    # the use of secrets in shared-workflows is discouraged, especially for public repositories.
-    # these values were added for situations where the use of secrets is unavoidable.
-    secrets:
-      RAPIDS_AUX_SECRET_1:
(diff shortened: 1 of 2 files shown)


gforsyth (approved):

trxcllnt:
Yes they have. The only use is in an inactive private repo that has an open PR to update.

PR #320, opened 2025-03-19
Add update-this-latest-branch workflow for this repo

KyleFromNVIDIA:


KyleFromNVIDIA:
We ended up tabling it because there's no way to update protected branches from a workflow. We would have to write a GitHub app. I think we can close it.

PR #321, opened 2025-03-19
Changed wheel_type to package_type

VenkateshJaya:
Changing the input parameter wheel_type to package_type for standardized usage across shared workflows, as `wheels-publish.yaml` uses package-type for the same purpose.

### DIFF
--- a/.github/workflows/wheels-build.yaml
+++ b/.github/workflows/wheels-build.yaml
@@ -13,6 +13,7 @@ on:
       sha:
         type: string
       build_type:
+        description: "One of: [branch, nightly, pull-request]"
         required: true
         type: string
       script:
@@ -82,6 +83,8 @@ jobs:
     steps:
       - name: Compute Build Matrix
         id: compute-matrix
+        env:
+          MATRIX_FILTER: ${{ inputs.matrix_filter }}
         run: |
           set -eo pipefail
 
@@ -108,7 +111,7 @@ jobs:
 
           MATRIX="$(
             yq -n -o json 'env(MATRIX)' | \
-            jq -c '${{ inputs.matrix_filter }} | if (. | length) > 0 then {include: .} else "Error: Empty matrix\n" | halt_error(1) end'
+            jq -c "${MATRIX_FILTER} | if (. | length) > 0 then {include: .} else \"Error: Empty matrix\n\" | halt_error(1) end"
           )"
 
           echo "MATRIX=${MATRIX}" | tee --append "${GITHUB_OUTPUT}"
--- a/.github/workflows/wheels-build.yaml
+++ b/.github/workflows/wheels-build.yaml
@@ -190,19 +190,19 @@ jobs:
       - name: Get package name
         if: inputs.package-type != ''
         env:
-          PACKAGE-TYPE: ${{ inputs.package-type }}
-          WHEEL-NAME: ${{ inputs.wheel-name }}
-          PURE-WHEEL: ${{ inputs.pure-wheel }}
+          PACKAGE_TYPE: ${{ inputs.package-type }}
+          WHEEL_NAME: ${{ inputs.wheel-name }}
+          PURE_WHEEL: ${{ inputs.pure-wheel }}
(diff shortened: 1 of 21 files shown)


.github/workflows/wheels-build.yaml
```diff
@@ -19,7 +19,7 @@ on:
       script:
         required: true
         type: string
-      wheel_type:
+      package_type:
```
jameslamb, line 22:
```suggestion
      package-type:
```

Could we use `package-type`, to make it exactly match the `wheels-publish` workflow?

https://github.com/rapidsai/shared-workflows/blob/51e5d508a8704ff11227af71d276e5b7c4d60d6b/.github/workflows/wheels-publish.yaml#L24-L28

I know it seems like a small thing, but being consistent here makes it clearer that these 2 inputs really do refer to the exact same concept, with the exact same set of valid values.

ajschmidt8 (approved):

jameslamb (changes requested):
Thanks for considering this! I left one request.

By the way, I noticed this PR has 33 commits in it: https://github.com/rapidsai/shared-workflows/pull/321/commits

I strongly suspect you're re-using a branch from a different PR. We squash all commits down into 1 on merge in this repo (and most RAPIDS repos), which is why `git` / GitHub are showing that you have 33 commits not on the target branch.

For the future, I recommend getting in the habit of updating your local checkout of the target branch and creating a clean new PR branch from that. Here's how I do that:

```shell
# just done one time at setup of my dev environment
git remote add upstream git@github.com:rapidsai/shared-workflows.git

# when I start a new PR
git checkout branch-25.04
git pull upstream branch-25.04
git checkout -b some-pr-branch
```

.github/workflows/wheels-build.yaml
VenkateshJaya: There's an input called `build_type` within this file, and seems to be a bit problematic to have a hypenated input along with an underscored input in the file. Let me know how do you want to resolve this.

jameslamb: That difference in casing already exists throughout these workflows.

For example:

https://github.com/rapidsai/shared-workflows/blob/51e5d508a8704ff11227af71d276e5b7c4d60d6b/.github/workflows/wheels-test.yaml#L27-L33

I wouldn't worry about it. I think that's less problematic than having two inputs whose names differ **only** by case.

.github/workflows/wheels-build.yaml
```diff
@@ -188,32 +188,32 @@ jobs:
...
       - name: Get package name
-        if: inputs.wheel_type != ''
+        if: inputs.package-type != ''
         env:
-          WHEEL_TYPE: ${{ inputs.wheel_type }}
-          WHEEL_NAME: ${{ inputs.wheel_name }}
-          PURE_WHEEL: ${{ inputs.pure_wheel }}
+          PACKAGE-TYPE: ${{ inputs.package-type }}
+          WHEEL-NAME: ${{ inputs.wheel-name }}
+          PURE-WHEEL: ${{ inputs.pure-wheel }}
```
ajschmidt8, line 195:
```suggestion
          PACKAGE_TYPE: ${{ inputs.package-type }}
          WHEEL_NAME: ${{ inputs.wheel-name }}
          PURE_WHEEL: ${{ inputs.pure-wheel }}
```

I'm not sure that `-` characters are valid for environment variable names.

When I ran `TEST-VAR=1` in `bash`, I got an error.

Can we switch the environment variables to use underscores? I realize this will cause them to be inconsistent with the `input` key names.

jameslamb: I agree, the environment variable form should have underscores.

VenkateshJaya: Oh I didn't know that. Changed the environment variables to use underscores now.

jameslamb (approved):
Thanks for considering my suggestions, looks good to me.

ajschmidt8 (approved):

PR #324, opened 2025-03-31
[WIP] Add token for private repo Github access

VenkateshJaya:
This PR adds functionality to control and validate the use of private GitHub tokens to download build artifacts from private repositories. It also includes changes to pr-builder to prevent PRs from being merged if they use the token to download artifacts from private repositories

## Implementation Details
- Each workflow now accepts an optional string input `artifact-token-name` (defaults to empty string)
- When set, the workflow uses the specified secret name to access private repositories instead of default `github.token`
  - The secret should be set on the repository calling the shared-workflows file
  - Usually inherited from the pr.yaml repository of the caller repository
- Token fallback mechanism using `${ARTIFACT_TOKEN:-$GH_TOKEN}` ensures default token is used when no custom token is provided
- The pr-builder workflow inspects all jobs' inputs to detect private token usage
- PRs using private tokens (indicated by setting `artifact-token-name`) will fail with the message: "PR validation failed: Private token access is not allowed to be merged onto the development branch"

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -22,10 +22,9 @@ on:
       matrix_filter:
         type: string
         default: "."
-      use-private-token:
+      artifact-token-name:
         required: false
-        type: boolean
-        default: false
+        type: string
 
 defaults:
   run:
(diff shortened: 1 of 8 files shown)


trxcllnt (changes requested):
~Typically we handle this with repo deploy keys. Is there a reason it needs to be a GH token instead?~

Nevermind, I just saw this is about artifact downloads. Not sure if there's a better way than GH tokens, but maybe there's a GitHub app we could use instead?

ajschmidt8:
> ~Typically we handle this with repo deploy keys. Is there a reason it needs to be a GH token instead?~
> 
> Nevermind, I just saw this is about artifact downloads. Not sure if there's a better way than GH tokens, maybe there's a GitHub app we could use instead?

From our conversations with Ray, we decided GH tokens were easier to manage.

PR #325, opened 2025-04-02
Enable proxy cache on NVKS V100 nodes.

bdice:
NVKS is migrating V100 nodes. Once this is complete, we can add proxy cache support to V100 jobs.


bdice:
@ajschmidt8 @jjacobelli Please give this a green light when the V100 jobs are all on NVKS.

PR #327, opened 2025-04-15
added descriptions to new variables in wheel build

VenkateshJaya:
Added descriptions to new variables in `wheels-build.yaml` explaining the function of those variables

### DIFF
--- a/.github/workflows/wheels-build.yaml
+++ b/.github/workflows/wheels-build.yaml
@@ -20,19 +20,23 @@ on:
         required: true
         type: string
       package-type:
+        description: "One of: [cpp, python]"
         required: false
         type: string
       wheel-name:
         required: false
         type: string
+        description: "The name of the wheel"
       pure-wheel:
         required: false
         type: boolean
         default: false
+        description: "Boolean flag to indicate if the wheel is a pure wheel"
       append-cuda-suffix:
         required: false
         type: boolean
         default: true
+        description: "Boolean flag to indicate if CUDA version should be appended to the wheel name"
 
       # allow a bigger runner instance
       node_type:


.github/workflows/wheels-build.yaml
```diff
@@ -20,19 +20,23 @@ on:
...
         type: string
       wheel-name:
         required: false
         type: string
+        description: "The name of the wheel"
       pure-wheel:
         required: false
         type: boolean
         default: false
+        description: "One of [true, false] to indicate if the wheel is a pure wheel"
```
jameslamb, line 34:
```suggestion
        description: "One of [true, false], true if the wheel is not dependent on operating system, Python version, or CPU architecture"
```

For someone who was already unsure about the name `pure-wheel`, I think that saying "is a pure wheel" in the description wouldn't help them to make a choice here. Would you consider this alternative wording?

.github/workflows/wheels-build.yaml
```diff
@@ -20,19 +20,23 @@ on:
         required: true
         type: string
       package-type:
+        description: "One of: [cpp, python]"
         required: false
         type: string
       wheel-name:
         required: false
         type: string
+        description: "The name of the wheel"
```
jameslamb, line 29:
```suggestion
        description: "Distribution name, without any other qualifiers (e.g. 'pylibcudf', not 'pylibcudf-cu12-cp311-manylinux_2_24_aarch64')"
```

If I was already unsure what to fill in for `wheel-name`, I don't think this description as-is would be much help to me. Would you please consider this rewording?

jameslamb (changes requested):
Sorry, just realized I made a mistake in https://github.com/rapidsai/cudf/pull/18346#discussion_r2042225162 ... I meant to say that any of these inputs without defaults should be `required: true`. To make it clear that a value must be provided.

I just edited that comment.

Could you please make that change here too?

jameslamb (approved):
Approving, the reason for not changing these to `required: true` is described in https://github.com/rapidsai/cudf/pull/18346#discussion_r2045604241

ajschmidt8 (approved):

PR #328, opened 2025-04-16
add mkdir -p to devcontainer setup

msarahan:
Lots of devcontainers jobs have been silently failing because this folder did not exist. We never properly set up devcontainers for telemetry. In https://github.com/rapidsai/rmm/pull/1883, @bdice had recently set bash error behavior flags, adding pipefail. That made this silent failure into a cryptic, blocking error.

This is set here and not in `rapids-telemetry-record` because we do not have gha-tools at this point.

### DIFF
--- a/.github/workflows/build-in-devcontainer.yaml
+++ b/.github/workflows/build-in-devcontainer.yaml
@@ -79,7 +79,7 @@ jobs:
 
       - name: Check if repo has devcontainer
         run: |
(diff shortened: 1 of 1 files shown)


bdice (approved):

msarahan:
test of this change is in https://github.com/rapidsai/rmm/pull/1892

msarahan:
Devcontainer builds in test pass, so I'll merge this and clean up RMM accordingly.

PR #329, opened 2025-04-17
Add `env` and `timeout-minutes` inputs to build-in-devcontainer.yaml

trxcllnt:


.github/workflows/build-in-devcontainer.yaml
```diff
@@ -14,6 +16,9 @@ on:
         default: '["conda", "pip"]'
       repo:
         type: string
+      timeout-minutes:
+        type: number
+        default: 360
```
trxcllnt, line 21: 6hrs is the current default, but this allows the caller to increase it.

bdice (approved):

PR #331, opened 2025-04-28
Add 'build_workflow_name' input, to allow overriding where scripts look for artifacts

VenkateshJaya:
Uniquely identifying an artifact from the GitHub Actions artifact store requires identifying the workflow run that produced it.

In PR CI, this is trivial because the uploading-artifacts and downloading-artifacts jobs are all running as part of the same workflow.

In other setups though, that isn't true. For example, RAPIDS uses 2 separate nightly workloads:

* `build.yaml` = build packages and publish the package artifacts
* `test.yaml` = pull those artifacts and run tests with them

For this case, in the relevant `gha-tools` scripts we default to assuming that `branch` or `nightly` runs want to pull artifacts from runs of a workflow literally called `build.yaml`:

```shell
run_id=$(
  RAPIDS_RETRY_SLEEP=30 \
    rapids-retry --quiet gh run list \
      ... \
      --workflow "${RAPIDS_BUILD_WORKFLOW_NAME:-build.yaml}" \
      ...
)
```

ref: https://github.com/rapidsai/gha-tools/blob/570fc211e6bf1bfb91098aa1bc4823ef7bb71fc2/tools/rapids-github-run-id#L66

There might be cases in the future where that assumption doesn't hold, and people want to pull artifacts in one workflow run from some other workflow which is NOT called `build.yaml`.

This PR + https://github.com/rapidsai/shared-actions/pull/52 add support for specifying the name of that "some other workflow" as an input, overriding the `build.yaml` default.

For more context, see https://github.com/rapidsai/gha-tools/pull/162#issuecomment-2831561646

## How I tested this

Tested with an `rmm` PR: https://github.com/rapidsai/rmm/pull/1909

* does not break existing repos CI: https://github.com/rapidsai/rmm/pull/1909#issuecomment-2917107512
* correctly threads this input through to the `RAPIDS_BUILD_WORKFLOW_NAME` variable: https://github.com/rapidsai/rmm/pull/1909#issuecomment-2917186610

### DIFF
--- a/.github/workflows/conda-cpp-post-build-checks.yaml
+++ b/.github/workflows/conda-cpp-post-build-checks.yaml
@@ -71,7 +71,7 @@ jobs:
           echo "RAPIDS_NIGHTLY_DATE=${{ inputs.date }}" >> "${GITHUB_ENV}"
       - name: Download conda C++ build artifacts
         run: |
-          CPP_DIR=$(rapids-download-conda-from-s3 cpp)
+          CPP_DIR=$(rapids-download-conda-from-github cpp)
           EXTRACTED_DIR=$(rapids-extract-conda-files "${CPP_DIR}")
           echo "RAPIDS_EXTRACTED_DIR=${EXTRACTED_DIR}" >> "${GITHUB_ENV}"
       - name: Get weak detection tool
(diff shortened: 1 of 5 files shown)


ajschmidt8 (approved):

jameslamb:
I've added context for why we're doing this in the description, and re-targeted this at `branch-25.08`. I need to run right now, but when I return to this I'll get it updated and merge it.

ajschmidt8 (approved):

PR #334, opened 2025-05-02
Remove deprecated wheel-name variable and make inputs required

VenkateshJaya:
This PR is part of the changes to move Build Artifact storage from S3 (`downloads.rapids.ai`) to the Github Artifact Store (see https://github.com/rapidsai/build-infra/issues/237). 

This PR removes the deprecated `wheel-name` variable (also refer https://github.com/rapidsai/shared-workflows/pull/330), and makes optional inputs required as changes have been made across RAPIDS repos.

PR #348, opened 2025-05-13
make input validation, upload logic in build workflows stricter

jameslamb:
Replaces #334

Contributes to https://github.com/rapidsai/build-infra/issues/237

Makes the following changes in the `wheels-build` workflow:

* removes input `wheel-name`
* makes input `package-name` required
* configures `actions/upload-artifact` to **ERROR** if no files are found to upload

### Why, and why now?

When the work to transition artifact-handling from `downloads.rapids.ai` to GitHub Actions artifacts started, this input referring to the name of the project was called `wheel-name`.

We later decided that `package-name` was more appropriate, for consistency with other workflows:

https://github.com/rapidsai/shared-workflows/blob/1ddf09797c78c158f23af3d4ef8a0df5801fcd07/.github/workflows/wheels-publish.yaml#L21

https://github.com/rapidsai/shared-workflows/blob/1ddf09797c78c158f23af3d4ef8a0df5801fcd07/.github/workflows/conda-cpp-build.yaml#L137

But by that time, some PRs calling it `wheel-name` had already been merged downstream. So both names were kept around for a bit.

Now, there are 0 remaining uses of `wheel-name` across the `rapidsai` org: https://github.com/search?q=org%3Arapidsai+%22wheel-name%22&type=code

### How I tested this

Searched for `wheel-name` uses across the `rapidsai` org and found none remaining: https://github.com/search?q=org%3Arapidsai+%22wheel-name%22&type=code

Searched for `wheels-build.yaml` across the `rapidsai` org and found that every use of it was also explicitly supplying a value for input `package-name:`: https://github.com/search?q=org%3Arapidsai+%22wheels-build.yaml%22+AND+NOT+is%3Aarchived&type=code

Except these repos which are not being actively maintained, and which we've agreed not to worry about for this migration:

* https://github.com/rapidsai/cuspatial/blob/c0f92e5b95f21ccc58d187689bb06adb2fd2bc56/.github/workflows/pr.yaml#L155

### DIFF
--- a/.github/workflows/wheels-build.yaml
+++ b/.github/workflows/wheels-build.yaml
@@ -20,17 +20,13 @@ on:
         required: true
         type: string
       package-name:
-        required: false
+        required: true
         type: string
         description: "Distribution name, without any other qualifiers (e.g. 'pylibcudf', not 'pylibcudf-cu12-cp311-manylinux_2_24_aarch64')"
       package-type:
         description: "One of: [cpp, python]"
         required: true
         type: string
-      wheel-name:
-        required: false
-        type: string
-        description: "DEPRECATED: Use package-name instead"
       pure-wheel:
         required: false
         type: boolean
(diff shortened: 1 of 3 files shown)


.github/workflows/wheels-build.yaml
```diff
@@ -241,6 +237,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         if: ${{ inputs.upload-artifacts }}
         with:
+          if-no-files-found: 'error'
```
jameslamb, line 240: Reference for this: https://github.com/actions/upload-artifact?tab=readme-ov-file#inputs

When `actions/upload-artifact` does not find any files to upload at `{path}`, by default it *warns* but does not *error*.

I think for our purposes, we want a big loud CI-failing error, as 0 files being found likely points to a misconfiguration. For any purposes where this workflow is run and you don't want to upload artifacts, we already have an opt-out via the `upload-artifacts` boolean input: https://github.com/rapidsai/shared-workflows/pull/340

ajschmidt8 (approved):

PR #339, opened 2025-05-07
Add CUDA 12.9.0

bdice:
Builds on top of https://github.com/rapidsai/shared-workflows/pull/268, as we are following the Python 3.13 migration with the CUDA 12.9.0 migration.

Contributes to https://github.com/rapidsai/build-planning/issues/173.

PR #268, opened 2024-11-19
Add Python 3.13

bdice:
Contributes to https://github.com/rapidsai/build-planning/issues/120.

This PR adds Python 3.13 entries to conda/wheel build and test jobs.

To keep the total number of GPU test jobs fixed, I modified the Python versions used by some existing jobs.

See previous migration to Python 3.12: https://github.com/rapidsai/shared-workflows/pull/213

PR #350, opened 2025-05-14
Replace all `-F` usage with `-f`

jarmak-nv:
This pull request updates multiple GitHub Actions workflows to standardize the usage of the `gh api graphql` command by replacing the `-F` flag with the `-f` flag for all GraphQL variable inputs. This change ensures consistency and aligns with best practices for the `gh` CLI.

From GitHub

> [Using variables](https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/using-the-api-to-manage-projects?#using-variables)
> In all of the following examples, you can use variables to simplify your scripts. **Use -F to pass a variable that is a number, Boolean, or null. Use -f for other variables.** For example,
> ```shell
> my_org="octo-org"
> my_num=5
> gh api graphql -f query='
>   query($organization: String! $number: Int!){
>     organization(login: $organization){
>       projectV2(number: $number) {
>         id
>       }
>     }
>   }' -f organization=$my_org -F number=$my_num
> ```
> For more information, see [Forming calls with GraphQL](https://docs.github.com/en/graphql/guides/forming-calls-with-graphql#working-with-variables).

I'm unsure if this is a recent change, but [we hit the failure](https://github.com/rapidsai/cudf/actions/runs/14986563348/job/42101512121#step:3:33) when attempting to convert an integer into a string. GitHub may silently fix string-likes.

targeting python3.13 as cuDF found the failure; can merge to main after confirmation of resolution.

### DIFF
--- a/.github/workflows/project-get-item-id.yaml
+++ b/.github/workflows/project-get-item-id.yaml
@@ -50,8 +50,12 @@ jobs:
             # Query up to 10 projects for the PR
             # There's no graphQL filter configured to query by a specific project
             # So we need to query all projects and filter the result ourselves
+<<<<<<< HEAD
             # shellcheck disable=SC2016
             gh api graphql -F nodeId="$ENV_ITEM_NODE_ID" -f query='
+=======
+            gh api graphql -f nodeId="$ENV_ITEM_NODE_ID" -f query='
+>>>>>>> 861fe12 (Replace all `-F` usage with `-f`)
               query($nodeId: ID!) {
                 node(id: $nodeId) {
                   ... on PullRequest {
--- a/.github/workflows/project-get-item-id.yaml
+++ b/.github/workflows/project-get-item-id.yaml
@@ -50,12 +50,7 @@ jobs:
             # Query up to 10 projects for the PR
             # There's no graphQL filter configured to query by a specific project
             # So we need to query all projects and filter the result ourselves
-<<<<<<< HEAD
-            # shellcheck disable=SC2016
-            gh api graphql -F nodeId="$ENV_ITEM_NODE_ID" -f query='
-=======
             gh api graphql -f nodeId="$ENV_ITEM_NODE_ID" -f query='
->>>>>>> 861fe12 (Replace all `-F` usage with `-f`)
               query($nodeId: ID!) {
                 node(id: $nodeId) {
                   ... on PullRequest {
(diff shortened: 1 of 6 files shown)


jameslamb (approved):

jameslamb:
@jarmak-nv before we merge... I noticed this was targeted at the `python-3.13` branch instead of `branch-25.06`... I suspect that's a mistake.

It does looks like there are some lingering references to the `@python-3.13` branch ([GitHub Search](https://github.com/search?q=org%3Arapidsai%20%22%40python-3.13%22&type=code)), but the PR for it (#268) has already been merged. I don't think we're planning to merge any new commits from the `python-3.13` back into `branch-25.06` at this point.

I've retargeted this at `branch-25.06`. I tried to do a rebase and force push, but I can't do that because this came from your fork. Could you please:

1. rebase this on the latest `branch-25.06`
2. force-push those changes to this branch

And then I think the deployment story should be:

1. merge this into `branch-25.06`
2. merge `branch-25.06` into #339
3. these changes will be out on all the repos `branch-25.06` branches once the CUDA 12.9 migration is done and they're switched back to `branch-25.06` (cc @gforsyth @bdice )

vyasr:
I think there's some confusion in part because it's going to be a bit before all the repos are using the same branch of shared-workflows anyway at this point, and I'm the one asking Ben for these workflow fixes that are blocking cudf's board automations (I don't think any other repo is using them, or at least not to the same extent) but cudf is lagging in some of the migrations due to e.g. the CUDA 12.9 compiler issues. Let's stick with James's suggestion for the merge plan. Worst case we need to merge down branch-25.06 into one of the migration branch sooner for more testing.

jarmak-nv:
Should be g2g

Will note cuML also uses these, currently pinned to branch-25.06.

jarmak-nv:
@jameslamb - when you're on, any thoughts on this CI failure?

```
Lint GitHub Actions workflow files.......................................Failed
- hook id: actionlint-docker
- exit code: 1

Executable `docker` not found
```

Pre-commit linting passes on my end, but maybe I broke something? :sweat_smile:

gforsyth:
@jarmak-nv -- this looks like the recently added `actionlint-docker` check doesn't work on `pre-commit.ci` -- it's also broken on `main`, so I don't think you broke anything.

gforsyth:
Ok, disabled it for CI only, so it'll still run locally. Going to merge this now

gforsyth:
Thanks @jarmak-nv !

jarmak-nv:
Thanks all! Appreciate the support :rocket:

PR #366, opened 2025-05-23
WIP: print GitHub API rate limit information to logs

jameslamb:
Contributes to https://github.com/rapidsai/build-infra/issues/241

Proposes printing information about GitHub API rate limits to logs.

### Intentionally targeting 25.06

We have a bit of a way to go on the 25.06 release, and I think we'll want this rate-limiting information available if more issues come up while trying to get the release out.

### How I tested this

TBD


ajschmidt8 (approved):
nice. glad to see that rate limits are 15,000 per hour like we'd expect. I find it hard to believe that we might have exceeded that even without your improvements. I guess we'll see how things look throughout the work day after your PR is merged.

PR #370, opened 2025-05-29
WIP: [NOT READY FOR REVIEW] Allow use of a different GitHub token from repo secrets

jameslamb:
Replaces #324

Occasionally, we want to run workflows which need more permissions for the GitHub API than the GitHub token automatically generated in each repo.

See https://github.com/rapidsai/build-infra/issues/242 (private issue, sorry)

This proposes adding an input to the workflows where RAPIDS has known use-cases for that type of thing, which allows specifying, at the call site for a workflow, the name of a secret in the repo which contains an alternative GitHub API token.

It also proposes adding logic to `pr-builder` to prevent checking in code that references that secret permanently. This mechanism is only intended for the type of testing described in https://docs.rapids.ai/resources/github-actions/#downloading-ci-artifacts.

### How I tested this

TBD

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -28,6 +28,13 @@ on:
       matrix_filter:
         type: string
         default: "."
+      alternative-gh-token-secret-name:
+        type: string
+        required: false
+        description: |
+          If provided, should contain the name of a secret in the repo which holds a GitHub API token.
+          When this is non-empty, that secret's value is used in place of the default repo-level token
+          anywhere that environment variable GH_TOKEN is set.
 
 defaults:
   run:
@@ -141,12 +148,12 @@ jobs:
               gh api /rate_limit | jq .
           fi
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ inputs.alternative-gh-token-secret-name && secrets[inputs.alternative-gh-token-secret-name] || github.token }}
       - name: C++ build
         run: ${{ inputs.script }}
         env:
           STEP_NAME: "C++ build"
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ inputs.alternative-gh-token-secret-name && secrets[inputs.alternative-gh-token-secret-name] || github.token }}
       - name: Get Package Name and Location
         if: ${{ inputs.upload-artifacts }}
         run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/pr-builder.yaml
```diff
@@ -15,3 +15,15 @@ jobs:
...
+        env:
+          NEEDS: ${{ inputs.needs }}
+          ERROR_MSG: "PR validation failed: Private token access is not allowed to be merged onto the development branch"
+        run: |
+          if jq -en 'env.NEEDS | fromjson | to_entries[] | select(.value.inputs.alternative-gh-token-secret-name and .value.inputs.alternative-gh-token-secret-name != "") | length > 0'; then
+            echo "::error::$ERROR_MSG"
+            echo "Jobs using private token:"
+            jq -en 'env.NEEDS | fromjson | to_entries[] | select(.value.inputs.alternative-gh-token-secret-name and .value.inputs.alternative-gh-token-secret-name != "") | "- " + .key'
+            exit 1
+          fi
```
jameslamb, line 29: As written, this is not working.

Fails like this:

```text
jq: error: gh/0 is not defined at <top-level>, line 1:
env.NEEDS | fromjson | to_entries[] | select(.value.inputs.alternative-gh-token-secret-name and .value.inputs.alternative-gh-token-secret-name != "") | length > 0  
```

([build link](https://github.com/rapidsai/cuml/actions/runs/15334190286/job/43149533223?pr=6809))

But then is "successful" (exit code 0) from the perspective of GitHub Actions, causing `pr-builder` to be ✅  even when you've set a value for `.values.inputs.alternative-gh-token-secret-name`.

Looking into it.

jameslamb: It looks like the `needs` context only contains workflow names, results, and outputs (exactly what you see in that screenshot above) by design: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context

jameslamb: @ajschmidt8 and I talked offline about this and he recommended that we could just checkout the repo's source code and search the files in `.github/` for uses of the token... I like that a lot 😁 

Pushed commits doing that: https://github.com/rapidsai/shared-workflows/pull/370/commits/618a4f3b0ec989b5ac1ba84d7d955d0f94855018 and https://github.com/rapidsai/shared-workflows/pull/370/commits/181f464aef36ed993ca81bb070968d93d178dd09

PR #372, opened 2025-05-30
obviate telemetry download

msarahan:
See https://github.com/rapidsai/shared-actions/pull/55 for more description

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -34,7 +34,8 @@ on:
         description: |
           If provided, should contain the name of a secret in the repo which holds a GitHub API token.
           When this is non-empty, that secret's value is used in place of the default repo-level token
-          anywhere that environment variable GH_TOKEN is set.
+          anywhere that environment variable GH_TOKEN is set. This is especially useful for downloading
+          artifacts from other private repos, which repo tokens do not have access to.
 
 defaults:
   run:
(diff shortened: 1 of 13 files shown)


.github/workflows/conda-cpp-build.yaml
```diff
@@ -132,9 +132,12 @@ jobs:
       - name: Telemetry setup
         uses: rapidsai/shared-actions/telemetry-dispatch-setup@main
         continue-on-error: true
-        if: ${{ vars.TELEMETRY_ENABLED == 'true' && github.run_attempt == '1' }}
+        if: ${{ vars.TELEMETRY_ENABLED == 'true' }}
+        env:
+          GH_TOKEN: ${{ inputs.alternative-gh-token-secret-name && secrets[inputs.alternative-gh-token-secret-name] || github.token }}
         with:
-            extra_attributes: "rapids.PACKAGER=conda,rapids.CUDA_VER=${{ matrix.CUDA_VER }},rapids.PY_VER=${{ matrix.PY_VER }},rapids.ARCH=${{ matrix.ARCH }},rapids.LINUX_VER=${{ matrix.LINUX_VER }}"
+          extra_attributes: "rapids.PACKAGER=conda,rapids.CUDA_VER=${{ matrix.CUDA_VER }},rapids.PY_VER=${{ matrix.PY_VER }},rapids.ARCH=${{ matrix.ARCH }},rapids.LINUX_VER=${{ matrix.LINUX_VER }}"
```
msarahan, line 139: Nothing changing here - just whitespace cleanup

.github/workflows/conda-cpp-build.yaml
```diff
@@ -132,9 +132,12 @@ jobs:
       - name: Telemetry setup
         uses: rapidsai/shared-actions/telemetry-dispatch-setup@main
         continue-on-error: true
-        if: ${{ vars.TELEMETRY_ENABLED == 'true' && github.run_attempt == '1' }}
+        if: ${{ vars.TELEMETRY_ENABLED == 'true' }}
+        env:
+          GH_TOKEN: ${{ inputs.alternative-gh-token-secret-name && secrets[inputs.alternative-gh-token-secret-name] || github.token }}
```
jameslamb, line 137:
```suggestion
          GH_TOKEN: ${{ github.token }}
```

Just putting this in writing from our offline conversation, so it's here in the history... all of these should just use `${{ github.token }}`. The telemetry stuff only needs to deal with artifacts from the same repo the workflow is running in, so the default repo-scoped token should be sufficient.

jameslamb (approved):
Looks good to me, based on our offline conversation and my reading of https://github.com/rapidsai/shared-actions/pull/55

Left one small suggestion, otherwise I think this is good to merge!

PR #374, opened 2025-06-03
create workflow to sync main branch to branch-25.08

msarahan:
This is one possible solution to https://github.com/rapidsai/ops/issues/3935#issuecomment-2851460675

This strategy makes branch-25.08 a one-way mirror of main. I used the --ff-only merge strategy so that things just fail if there's any kind of ambiguity in the merge.

There are other ways to do this that create PRs instead of direct pushes, such as https://github.com/marketplace/actions/sync-branches. I'm not sure if we need them.

### DIFF
--- a/.github/workflows/branch-migration-sync.yaml
+++ b/.github/workflows/branch-migration-sync.yaml
@@ -1,0 +1,22 @@
+name: Sync main to branch-25.08 (migration)
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


msarahan:
Ray says this isn't needed. We'll just change the branch-25.08 (or whatever's current) to main and adapt all consuming repos at once.

PR #390, opened 2025-06-30
feat: add zizmor static action checks

gforsyth:
Re-opening this PR instead of #351 so that the PR is coming from a local branch.

This is an experiment in using `zizmor` (https://docs.zizmor.sh/) to run static analysis on all of the shared workflow files.  This seems like a nice tool to add, especially in this repository.

Some of the classes of errors/warnings that are fixed here (and any associated configuration):

### unpinned-uses
https://docs.zizmor.sh/audits/#unpinned-uses

Any `uses:` mapping should pin to an exact hash to avoid things getting swapped out underneath.
Since we deliberately make use of some of the `uses:` to reference our own `shared-workflows` and `nv-gha-runners` -- these are allowed to point at refs.  Those patterns are configured in `.github/zizmor.yml`

### artipacked
https://docs.zizmor.sh/audits/#artipacked

We should always explicitly set `persist-credentials`, either to `true` or `false`.  Everywhere that the `checkout` action has implicitly set it to `true`, I have made it an explicit `true`.

I suspect some of these need to be true, but probably not all of them.  I'll have to do some testing downstream to see which actions break when those secrets are disabled.

### overprovisioned-secrets
https://docs.zizmor.sh/audits/#overprovisioned-secrets

Looking up secrets like `secrets[some_key]` injects the entire secrets context into the runner, but I don't see a way around that for our current secret lookup patterns. I've just ignored these instances individually.

### unpinned-images
https://docs.zizmor.sh/audits/#unpinned-images

The only unpinned images belong to us and those have been individually allowlisted.

### template-injection
https://docs.zizmor.sh/audits/#template-injection

Generally speaking lines like `run: ${{ inputs.script }}` are susceptible to template injection.  Also, that's the entire point of these lines, so they are individually allowlisted.

For other potentially injectable lines, like `run: rapids-wheels-anaconda "${{ inputs.package-name }}" "${{ inputs.package-type }}" `, I've unpacked the inputs in to the environment and passed through those env-vars instead.

In #351, @msarahan and @jameslamb and I determined that we can use full SHAs to specify the image to use for a given step, but leave a trailing same-line comment that points to the corresponding tag.

That let's us use the more secure exact pin for the images that we don't control, but still keeps it human-readable and compatible with dependabot, e.g.

```
- uses: actions/checkout@def456 # v4.2.2

```

### DIFF
--- a/.github/workflows/breaking-change-alert.yaml
+++ b/.github/workflows/breaking-change-alert.yaml
@@ -124,7 +124,7 @@ jobs:
           fi
 
       - name: Send Slack notification
-        uses: slackapi/slack-github-action@485a9d42d3a73031f12ec201c457e2162c45d02d
+        uses: slackapi/slack-github-action@485a9d42d3a73031f12ec201c457e2162c45d02d # v2.0.0
         with:
           payload: |
             {
--- a/.github/workflows/breaking-change-alert.yaml
+++ b/.github/workflows/breaking-change-alert.yaml
@@ -124,7 +124,7 @@ jobs:
           fi
 
       - name: Send Slack notification
-        uses: slackapi/slack-github-action@v2.0.0
+        uses: slackapi/slack-github-action@485a9d42d3a73031f12ec201c457e2162c45d02d
         with:
           payload: |
             {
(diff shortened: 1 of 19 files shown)


bdice (approved):

gforsyth:
Ok, I'm going to merge this in.  If everything in CI breaks, we know what to revert

AyodeAwe:
Cross-linking threads for posterity as this PR has no backing Github issue:
- https://nvidia.slack.com/archives/C06EYSLPGHX/p1752692068689939
- https://nvidia.slack.com/archives/C021UF8PNMN/p1741974322941259

PR #391, opened 2025-07-01
conda-cpp-post-build-checks: remove input 'enable_check_symbols'

### LINKED ISSUE #379: conda-cpp-post-build-checks: remove `check_symbols` input
## Description

`conda-cpp-post-build-checks` accepts an input, `enable_check_symbols`, which defaults to `false`.

https://github.com/rapidsai/shared-workflows/blob/2ab9090c740ccabee7f26dfa1405a2416289aece/.github/workflows/conda-cpp-post-build-checks.yaml#L16-L19

That input being `false` skips the only job defined in the workflow:

https://github.com/rapidsai/shared-workflows/blob/2ab9090c740ccabee7f26dfa1405a2416289aece/.github/workflows/conda-cpp-post-build-checks.yaml#L42-L44

Which means that **by default, this workflow does not do anything**.

That is probably why every current use of `conda-cpp-post-build-checks` is passing `enable_check_symbols: true` ([GitHub Search](https://github.com/search?q=org%3Arapidsai+%22conda-cpp-post-build-checks.yaml%22+AND+NOT+is%3Aarchived&type=code))

We should just remove that input.

## Benefits of this work

* removes unnecessary complexity... just like all other workflows, you could skip it simply by not ever calling it

## Acceptance Criteria

* `conda-cpp-post-build-checks` does not have input `enable_check_symbols`
* CI is working in all repos using `conda-cpp-post-build-checks`

## Approach

1. switch the default to for `enable_check_symbols` to `true` (#382)
2. remove all uses of that input in repos using the `conda-cpp-post-build-checks` workflow ([GitHub Search](https://github.com/search?q=org%3Arapidsai+%22conda-cpp-post-build-checks.yaml%22+AND+NOT+is%3Aarchived&type=code))
  - [x] `cudf` (https://github.com/rapidsai/cudf/pull/19234/)
  - [x] `cugraph` (https://github.com/rapidsai/cugraph/pull/5139)
  - [x] `cuml` (https://github.com/rapidsai/cuml/pull/6952)
  - [x] `cuvs` (https://github.com/rapidsai/cuvs/pull/1075)
  - [x] `raft` (https://github.com/rapidsai/raft/pull/2728)
3. remove `enable_check_symbols` input in `conda-cpp-post-build-checks` workflow

## Notes

Discovered this while working on documenting inputs (#376).

### Why was this input added at all?

I'm not sure. It's been there since the very beginning (#152), and since the very beginning there's only been a single job in this workflow.

I suspect that maybe it was following the pattern from `checks.yaml`, which has multiple jobs and boolean flags to turn each of them off.

https://github.com/rapidsai/shared-workflows/blob/2ab9090c740ccabee7f26dfa1405a2416289aece/.github/workflows/checks.yaml#L9-L28

If we don't want to remove this input, I think we at least should switch its default to `true` and remove it from all the downstream call sites.

jameslamb:
Closes #379 

This removes the input `enable_check_symbols` from the `conda-cpp-post-build-checks` workflow.

## Notes for Reviewers

Labeling this "breaking" because it would break any callsites that pass that input... but I believe I've already updated all of those (see PRs in #379), so **this should be safe to merge**.

### DIFF
--- a/.github/workflows/build-in-devcontainer.yaml
+++ b/.github/workflows/build-in-devcontainer.yaml
@@ -121,6 +121,7 @@ jobs:
           env: |
             REPOSITORY=${{ env.REPOSITORY }}
             SCCACHE_REGION=${{ vars.AWS_REGION }}
+            SCCACHE_ERROR_LOG=/home/coder/${{ env.REPOSITORY }}/sccache.log
             AWS_ACCESS_KEY_ID=${{ env.AWS_ACCESS_KEY_ID }}
             AWS_SESSION_TOKEN=${{ env.AWS_SESSION_TOKEN }}
             AWS_SECRET_ACCESS_KEY=${{ env.AWS_SECRET_ACCESS_KEY }}
@@ -133,13 +134,6 @@ jobs:
             ${{ inputs.env }}
           runCmd: |
             set -euo pipefail;
-            mkdir -p ~/.config/pip/;
-            cat <<EOF >> ~/.config/pip/pip.conf
-            [global]
-            extra-index-url = https://pypi.anaconda.org/rapidsai-wheels-nightly/simple
-            EOF
-
-            rapids-make-${PYTHON_PACKAGE_MANAGER}-env;
 
             if test -n '${{ inputs.extra-repo-deploy-key }}' \
             || test -n '${{ inputs.extra-repo-deploy-key-2 }}'; then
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
On the current `HEAD` I can confirm that the only references to `enable_check_symbols` appear in this file:

```sh
󰕈 gforsyth  …/shared-workflows   branch-25.08   11:01 
🐚 git rev-parse HEAD
86c81b17e9f198ff527d6b34f5adc2ce90ce7f5a

󰕈 gforsyth  …/shared-workflows   branch-25.08   11:01 
🐚 ag --hidden "enable_check_symbols"
.github/workflows/conda-cpp-post-build-checks.yaml
22:      enable_check_symbols:
55:    if: ${{ inputs.enable_check_symbols }}
```

Looking at the broader RAPIDS workflows (https://github.com/search?q=org%3Arapidsai+%22conda-cpp-post-build-checks.yaml%22+AND+NOT+is%3Aarchived&type=code)

I see `enable_check_symbols` being used in `cuspatial` (which isn't developed/published anymore) and in a private repo that hasn't been touched in 7 months (and is pointing at an older branch of `shared-workflows` anyway)

jameslamb:
Great, thanks for double-checking my search! Those are the same other uses I saw and decided do not need to be updated.

PR #393, opened 2025-07-16
custom-job: make artifact name customizable

### LINKED ISSUE #392: custom-job: Option to customize name of the artifact being uploaded in custom job workflow
Hi, Currently the name is result by default, is it possible to make this an optional input with default name "result", so users can choose the name for the artifact ?

https://github.com/rapidsai/shared-workflows/blob/4bf94447dd66683d8e33c4212dab35a6aa8e2cd8/.github/workflows/custom-job.yaml#L120

jameslamb:
Closes #392

The `custom-job` workflow allows you to invoke `actions/upload-artifact` by passing a path to a file / directory to the input `file_to_upload`.

The artifact name is hard-coded in the workflow...

https://github.com/rapidsai/shared-workflows/blob/a8fe84c8249ec3553f271fe8ed3b5247543b3561/.github/workflows/custom-job.yaml#L152-L155

...which mean **you cannot have 2 `custom-job` calls that upload artifacts in the same workflow run**.

This fixes that by making the artifact name configurable. It preserves the current hard-coded name (`"result"`) as a default, so no existing uses which happen to depend on that name will be broken.

### DIFF
--- a/.github/workflows/breaking-change-alert.yaml
+++ b/.github/workflows/breaking-change-alert.yaml
@@ -124,7 +124,7 @@ jobs:
           fi
 
       - name: Send Slack notification
-        uses: slackapi/slack-github-action@v2.0.0
+        uses: slackapi/slack-github-action@485a9d42d3a73031f12ec201c457e2162c45d02d # v2.0.0
         with:
           payload: |
             {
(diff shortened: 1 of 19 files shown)


gforsyth (approved):

rgsl888prabhu (approved):
Thank you @jameslamb for a quick PR

PR #396, opened 2025-07-18
ci: prevent renovate from incorrectly bumping rapidsai docker images

gforsyth:
Closes #394 
Closes #395 

Renovate is seeing our updated docker images for the 25.10 branch and thinking it should update the images here.  It should not do this, and the bump is handled by our own `update-version.sh` script, so I'm disabling renovate updates for the `rapidsai/` dockerhub org.


jameslamb:
Thanks for taking care of this @gforsyth !!!

PR #400, opened 2025-07-21
Always upload sccache logs in `build-in-devcontainer.yaml`

trxcllnt:
Upload sccache client logs on success and failure, except when the workflow is cancelled.

### DIFF
--- a/.github/workflows/build-in-devcontainer.yaml
+++ b/.github/workflows/build-in-devcontainer.yaml
@@ -192,7 +192,7 @@ jobs:
             cd ~/"${REPOSITORY}";
             mkdir -p telemetry-artifacts;
             ${{ inputs.build_command }}
-      - if: ${{ env.HAS_DEVCONTAINER == 'true' }}
+      - if: ${{ !cancelled() && env.HAS_DEVCONTAINER == 'true' }}
         name: Upload sccache logs
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-in-devcontainer.yaml
```diff
@@ -159,7 +159,11 @@ jobs:
             AWS_ACCESS_KEY_ID=${{ env.AWS_ACCESS_KEY_ID }}
             AWS_SESSION_TOKEN=${{ env.AWS_SESSION_TOKEN }}
             AWS_SECRET_ACCESS_KEY=${{ env.AWS_SECRET_ACCESS_KEY }}
-            RAPIDS_AUX_SECRET_1=${{ inputs.rapids-aux-secret-1 != '' && secrets[inputs.rapids-aux-secret-1] || '' }} # zizmor: ignore[overprovisioned-secrets]
+            RAPIDS_AUX_SECRET_1=${{
+              inputs.rapids-aux-secret-1 != ''
+              && secrets[inputs.rapids-aux-secret-1] # zizmor: ignore[overprovisioned-secrets]
+              || ''
+            }}
```
trxcllnt, line 166: The `env:` key for this action is a string field, not a YAML map.

The `# zizmor: ignore[overprovisioned-secrets]` comment here means the envvar is the literal value:
```
RAPIDS_AUX_SECRET_1="<secret> # zizmor: ignore[overprovisioned-secrets]"
```

PR #401, opened 2025-07-22
Pass `--quiet` flag in `build-in-devcontainer` workflow

TomAugspurger:
This passes the `--quiet` flag to the environment creation command, to avoid noisy output from progress bars like https://github.com/rapidsai/cudf/actions/runs/16429104215/job/46426604875?pr=19411#step:9:2240.


bdice:
Here are the historical issues we had with enabling `--quiet`:
https://github.com/rapidsai/ci-imgs/pull/220
https://github.com/rapidsai/ci-imgs/issues/228

PR #408, opened 2025-08-08
Cache sccache preprocessor and toolchain dirs across runs

trxcllnt:
This PR caches the `.cache/{sccache,sccache-dist-client}` dirs across CI runs so builds are accelerated by sccache's preprocessor cache mode.

### DIFF
--- a/.github/workflows/build-in-devcontainer.yaml
+++ b/.github/workflows/build-in-devcontainer.yaml
@@ -186,7 +186,7 @@ jobs:
           env: |
             REPOSITORY=${{ env.REPOSITORY }}
             SCCACHE_REGION=${{ vars.AWS_REGION }}
-            SCCACHE_ERROR_LOG=/home/coder/.cache/sccache.log
+            SCCACHE_ERROR_LOG=/home/coder/${{ env.REPOSITORY }}/sccache.log
             AWS_ACCESS_KEY_ID=${{ env.AWS_ACCESS_KEY_ID }}
             AWS_SESSION_TOKEN=${{ env.AWS_SESSION_TOKEN }}
             AWS_SECRET_ACCESS_KEY=${{ env.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-in-devcontainer.yaml
```diff
@@ -119,36 +120,67 @@ jobs:
...
+          RUN_ATTEMPT: ${{ github.run_attempt }}
         run: |
-          echo "REPOSITORY=$(basename "$(pwd)")" | tee -a "${GITHUB_ENV}"
-          if test -f ".devcontainer/cuda${CUDA_VER}-${PACKAGER}/devcontainer.json"; then
-            echo "HAS_DEVCONTAINER=true" >> "${GITHUB_ENV}"
-          else
-            echo "HAS_DEVCONTAINER=false" >> "${GITHUB_ENV}"
+          HAS_DEVCONTAINER=false
+          if test -f "repo/.devcontainer/cuda${CUDA_VER}-${PACKAGER}/devcontainer.json"; then
+            HAS_DEVCONTAINER=true
```
bdice, line 131: If this is false because there’s no devcontainer, what does the “build devcontainer” workflow do? How does the conditional help us?

jameslamb: > _If there's no devcontainer, we want to do nothing without failing the build._

Why would we not want to fail the build in that case?

I don't think we should do this... it could silently cover up actual problems. For example, if any repo deviates from the standard naming convention we have (`cuda${CUDA_VER}-${PACKAGER}/devcontainer.json`), today you get a big loud CI failure telling you about that.

With this check, that'd turn into this job just silently being a no-op, which might be missed.

Repos without devcontainers should just not use this workflow, and this workflow should fail outright if what it needs isn't found.

jameslamb: We talked offline about this... this is not new behavior, it's only showing up here because of the change to git cloning that now writes the repo to `repo/` (changing the file path passed to `test -f`).

Approving this, but I do think we should remove this `HAS_DEVCONTAINER=false` behavior in a follow-up.

bdice: Great. I agree we should fail loudly here, and remove this in a follow-up.

Issue #427, opened 2025-09-24
Shift some arm64 testing jobs to L4s?


jameslamb:
## Description

There are now arm64 L4 runners available in NVIDIA's self-hosted CI runners.

<img width="870" height="174" alt="Image" src="https://github.com/user-attachments/assets/4722f582-4341-420f-afd0-a5878f03f2f2" />

We should consider shifting some RAPIDS arm64 testing jobs (currently all done on A100s) there.

## Benefits of this work

* would relieve some contention for A100s (hopefully leading to less waiting in CI)
* would help nudge projects to ensure their tests are portable to different GPUs, available memory, etc.

## Acceptance Criteria

* some RAPIDS CI arm64 jobs use `linux-arm64-gpu-l4-{earliest,latest}-1` runners
* 0 net-new RAPIDS CI jobs

## Approach

Update the matrices in the relevant testing workflows (any using GPUs on arm64 hosts):

https://github.com/rapidsai/shared-workflows/blob/b8ea6d3e1f4d9a2f608633ae8c7482d40707a505/.github/workflows/wheels-test.yaml#L129-L147

## Notes

N/A

bdice:
Done as of #428.

PR #428, opened 2025-09-24
Use ARM L4 nodes

bdice:
See #427.

This aims to use ARM L4 nodes for most of our CI, while ensuring we retain coverage of sm80 (A100) on PRs and nightly tests.


jameslamb (approved):
I support the overall change in the mix, but it looks like we'd now have 0 test jobs (PR or nightly) with CUDA 13 + A100s.

Would you consider shifting one of these around so we get at least one CUDA 13 + A100 job somewhere in the matrix?

bdice:
All tests passed. Merging this. https://github.com/rapidsai/cudf/pull/20114#issuecomment-3336521364

PR #435, opened 2025-10-23
feat(wheels,conda): support passing `RAPIDS_PACKAGE_NAME` from build step to artifact upload

gforsyth:
This is in support of (part of) rapidsai/build-planning#204

There are other changes that are in `gha-tools` for how to configure the artifact names for uploading and downloading interim wheels and conda packages (rapidsai/gha-tools#218 rapidsai/gha-tools#219).

The change here is to allow setting the value of `RAPIDS_PACKAGE_NAME` in a downstream build script and then having that value passed to the `Get package name` step so that we have the option to configure artifact names on a per-job basis.

### DIFF
--- a/.github/workflows/wheels-build.yaml
+++ b/.github/workflows/wheels-build.yaml
@@ -305,7 +305,7 @@ jobs:
           if [[ -n "${RAPIDS_PACKAGE_NAME:-}" ]]; then
             echo "RAPIDS_PACKAGE_NAME=${RAPIDS_PACKAGE_NAME}" >> "${GITHUB_OUTPUT}"
           else
-            echo "RAPIDS_PACKAGE_NAME=$(RAPIDS_NO_PKG_EXTENSION=true rapids-package-name "wheel_${PACKAGE_TYPE}")" >> "${GITHUB_OUTPUT}"
+            echo "RAPIDS_PACKAGE_NAME=$(rapids-package-name "wheel_${PACKAGE_TYPE}")" >> "${GITHUB_OUTPUT}"
           fi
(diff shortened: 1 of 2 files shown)


.github/workflows/wheels-build.yaml
```diff
@@ -290,7 +300,13 @@ jobs:
           if [ "${PURE_WHEEL}" = "true" ]; then
             export "RAPIDS_PY_WHEEL_PURE=1"
           fi
-          echo "RAPIDS_PACKAGE_NAME=$(RAPIDS_NO_PKG_EXTENSION=true rapids-package-name "wheel_${PACKAGE_TYPE}")" >> "${GITHUB_OUTPUT}"
+
+          # Use RAPIDS_PACKAGE_NAME from build step if available, otherwise check environment, otherwise generate default
+          if [[ -n "${RAPIDS_PACKAGE_NAME:-}" ]]; then
+            echo "RAPIDS_PACKAGE_NAME=${RAPIDS_PACKAGE_NAME}" >> "${GITHUB_OUTPUT}"
+          else
+            echo "RAPIDS_PACKAGE_NAME=$(RAPIDS_NO_PKG_EXTENSION=true rapids-package-name "wheel_${PACKAGE_TYPE}")" >> "${GITHUB_OUTPUT}"
```
vyasr:
```suggestion
            echo "RAPIDS_PACKAGE_NAME=$(rapids-package-name "wheel_${PACKAGE_TYPE}")" >> "${GITHUB_OUTPUT}"
```

vyasr (approved):
Needs conflict resolution, and the two comments are probably related to the cause of conflict, but otherwise LGTM.

PR #438, opened 2025-10-30
Add v100 coverage for CUDA 12 nightlies

bdice:
V100 GPUs (sm70) are supported on CUDA 12. We should run nightlies on Volta for increased architecture coverage.


vyasr (approved):
I support checking our oldest supported hardware, at least in nightlies like in this PR.

PR #439, opened 2025-10-30
Add workflow inputs for enabling `sccache-dist`

trxcllnt:
This PR adds an `sccache-dist-token-secret-name` input to our conda and wheel workflows. When the name of a secret is provided, the workflow will invoke the [`rapidsai/shared-workflows/setup-sccache-dist`](https://github.com/rapidsai/shared-actions/pull/81) to install the latest sccache client from [`rapidsai/sccache`](https://github.com/rapidsai/sccache) and configure it for distributed compilation.

This PR also adds the sccache client logfile as an additional build artifact so the logs can be inspected if anything goes wrong.

Tested in https://github.com/rapidsai/rmm/pull/2101 (example workflow run [here](https://github.com/rapidsai/rmm/actions/runs/18859867864/job/53815921352?pr=2101)).

### DIFF
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -219,7 +219,7 @@ jobs:
 
       # Install latest rapidsai/sccache client and configure sccache-dist
       - name: Setup sccache-dist
-        uses: rapidsai/shared-actions/setup-sccache-dist@fea/setup-sccache-dist
+        uses: rapidsai/shared-actions/setup-sccache-dist@main
         if: ${{ inputs.sccache-dist-token-secret-name != '' }}
         env:
           AWS_REGION: "${{env.AWS_REGION}}"
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -169,7 +169,7 @@ jobs:
 
       # Install latest rapidsai/sccache client and configure sccache-dist
       - name: Setup sccache-dist
-        uses: rapidsai/shared-actions/setup-sccache-dist@fea/setup-sccache-dist
+        uses: rapidsai/shared-actions/setup-sccache-dist@main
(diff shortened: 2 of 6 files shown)


.github/workflows/conda-python-tests.yaml
```diff
@@ -206,6 +216,19 @@ jobs:
...
+      - name: Setup sccache-dist
+        uses: rapidsai/shared-actions/setup-sccache-dist@fea/setup-sccache-dist
+        if: ${{ inputs.sccache-dist-token-secret-name != '' }}
+        env:
+          AWS_REGION: "${{env.AWS_REGION}}"
+          AWS_ACCESS_KEY_ID: "${{env.AWS_ACCESS_KEY_ID}}"
+          AWS_SECRET_ACCESS_KEY: "${{env.AWS_SECRET_ACCESS_KEY}}"
+        with:
+          auth: "${{ secrets[inputs.sccache-dist-token-secret-name] }}" # zizmor: ignore[overprovisioned-secrets]
+          cache-slug: "conda-py${{matrix.PY_VER}}-cuda${{matrix.CUDA_VER}}-${{matrix.ARCH}}"
```
rockhowse: Is there no corresponding `conda-python-build` change that is used with the `conda-py` slug?

I see the `conda-cpp-build` and `wheels-build` have a `build` and a `test` componant, bu the `python` side is just `tests` 

And I see a similarly structured `Python build` block in `conda-python-build.yaml` here:

* https://github.com/rapidsai/shared-workflows/blob/main/.github/workflows/conda-python-build.yaml#L179-L183

But no changes that setup the sccache-dist with a `conda-py` based slug or being used in the aforementioned `Python build` section.

trxcllnt: Good point, I'll add it there too

bdice (approved):

PR #470, opened 2025-11-25
increase nofile ulimits

trxcllnt:
Increase nofile ulimits for sccache.


trxcllnt:
Tested in https://github.com/rapidsai/cudf/pull/20717

PR #487, opened 2026-01-07
build-in-devcontainer: fail if expected devcontainer directories aren't found

jameslamb:
Follow-up from https://github.com/rapidsai/shared-workflows/pull/408#discussion_r2267393154

As of this PR, if the `build-in-devcontainer` workflow is called and the calling repo is missing the expected devcontainers directory, it'll fail loudly instead of passing silently.

## Notes for Reviewers

The extra flexibility here bit us during https://github.com/rapidsai/build-planning/issues/236. In the first round of PRs there, RAPIDS devcontainer directories were renamed from like `cuda13.0-pip` to `cuda13.1-pip` but the corresponding call of this workflow wasn't updated. The devcontainer CI jobs passed in CI and we didn't notice that builds were not actually running.

As of this PR, they'd fail loudly and alert us to that.

### DIFF
--- a/.github/workflows/build-in-devcontainer.yaml
+++ b/.github/workflows/build-in-devcontainer.yaml
@@ -126,37 +126,28 @@ jobs:
           RUN_ID: ${{ github.run_id }}
           RUN_ATTEMPT: ${{ github.run_attempt }}
         run: |
-          HAS_DEVCONTAINER=false
-          if test -f "repo/.devcontainer/cuda${CUDA_VER}-${PACKAGER}/devcontainer.json"; then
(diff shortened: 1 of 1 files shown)


rockhowse (commented):
I don't 100% fully comprehend the world of `devcontainers` yet and defer the actually ✅ to someone more knowledgeable like @trxcllnt 

Once someone with more `devcontainer` related 👀 feel free to ping me and I am happy to ✅ as I in-general support fail fast and loud for critical things that were currently failing silently causing confusion and engineering time trying to track down.

PR #498, opened 2026-02-03
use CUDA 13.1.1

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/248

There are CUDA 13.1.1 CI images now (https://github.com/rapidsai/ci-imgs/pull/360), this proposes using them instead of CUDA 13.1.0 everywhere.

This is a patch release and should be relatively safe, so proposing just immediately merging this instead of doing a rollout with swapping branch references.


bdice:
Sounds good, I'll merge.

PR #500, opened 2026-02-04
Fix when there is no changed-files workflow

bdice:
I have a repo with no VERSION file. I set `enable_check_version_files_changed: false` in `checks.yaml` but still got an error:

```
Error: The template is not valid. rapidsai/shared-workflows/.github/workflows/checks.yaml@main (Line: 97, Col: 34): Error reading JToken from JsonReader. Path '', line 0, position 0.
```

CI logs: https://github.com/rapidsai/velox-testing/actions/runs/21686871425/job/62536173276

The issue is that this is expecting output from a `changed-files` job that doesn't exist. We can work around this by including a check for that case and fallback value.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -94,7 +94,7 @@ jobs:
             exit 1
           fi
         env:
-          VERSION_FILES_CHANGED: ${{ (steps.changed-files.outputs.changed_file_groups && fromJSON(steps.changed-files.outputs.changed_file_groups).version_files) || 'false' }}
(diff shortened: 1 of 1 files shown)


KyleFromNVIDIA:
Something isn't right. I guarded that step with the following condition specifically to prevent this kind of thing from happening:

```
if: ${{ inputs.enable_check_version_files_changed }}
```

Is GHA evaluating the `env` even if the `if` is false? If so, I'd be in favor of filing a bug with GHA.

KyleFromNVIDIA (changes requested):

KyleFromNVIDIA (approved):

KyleFromNVIDIA:
I've opened https://github.com/orgs/community/discussions/186382 to discuss short-circuit evaluation. Approving this PR to work around the issue.

PR #503, opened 2026-02-10
Use GHA id-token as `sccache-dist` auth token

trxcllnt:
Tested in https://github.com/rapidsai/devcontainers/pull/663 and https://github.com/rapidsai/rmm/pull/2249

**Merge after https://github.com/rapidsai/shared-actions/pull/95**


jameslamb (approved):
Great, after this can you please remove the `sccache-dist-token-secret-name` input in all the callsites around RAPIDS and then come back and remove it from these workflows?

PR #520, opened 2026-04-13
feat: add "bootstrap" option to `custom_job`

gforsyth:
While working on rapidsai/integration#837, I ran into an issue where we are specifically using lightweight `nvidia/cuda` images to test a clean install environment, but `shared-workflows` expects at least `git`, `jq`, and `curl` to be installed before the custom script is executed. (These dependencies are from the `rapids-github-info` and `setup-sccache` shared actions, respectively.

Adding an option here to run a smaller, bootstrap-only script to install any missing dependencies that would block the custom job from even making it to the custom script it is running.

Happy to discuss this further, but wanted to push this up to try it out.

### DIFF
--- a/.github/workflows/custom-job.yaml
+++ b/.github/workflows/custom-job.yaml
@@ -97,6 +97,17 @@ on:
         default: false
         type: boolean
         required: false
+      bootstrap_script:
+        description: |
+          A script to run immediately after pulling the requested container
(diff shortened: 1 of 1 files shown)


gforsyth:
Might be useful in another context, but insufficient for the work linked above.

PR #524, opened 2026-04-21
Split matrix computation into its own workflow

KyleFromNVIDIA:
This allows custom workflows to take advantage of these matrices, and avoid hard-coding CUDA versions into them.

Tested in https://github.com/rapidsai/cudf/pull/22240


gforsyth (approved):
Excellent -- I needed this a couple of days ago and now it's available!

PR #528, opened 2026-04-27
feat: add override env-var for setting custom artifact name

gforsyth:
This is part of rapidsai/build-planning#270.

I'm rolling out a new `rapids-artifact-name` tool to make our artifact names more consistent - part of the first phase of this is overriding the "default" names that get created by `shared-workflows`.  I previously added the capability to override the artifact name from within the ci scripts for the stable ABI work, but that only covered the `python-build` workflows, this extends that capability to the `cpp-build` workflows.

Tested this in https://github.com/rapidsai/rmm/pull/2370 along with the vendored version of the script from https://github.com/rapidsai/gha-tools/pull/253

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -169,9 +169,16 @@ jobs:
           # NEEDS alternative-gh-token-secret-name - API limits need to be for whatever token is used for upload/download. Repo token may be a different pool for rate limits.
           GH_TOKEN: ${{ inputs.alternative-gh-token-secret-name && secrets[inputs.alternative-gh-token-secret-name] || github.token }} # zizmor: ignore[overprovisioned-secrets]
       - name: C++ build
+        id: cpp-build
         run: |
           ulimit -n "$(ulimit -Hn)"
-          $INPUTS_SCRIPT
+          # shellcheck disable=SC1090
+          source "${INPUTS_SCRIPT}"
+
+          # Capture RAPIDS_PACKAGE_NAME if set by the build script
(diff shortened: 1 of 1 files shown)


.github/workflows/conda-cpp-build.yaml
```diff
@@ -169,18 +169,33 @@ jobs:
           # NEEDS alternative-gh-token-secret-name - API limits need to be for whatever token is used for upload/download. Repo token may be a different pool for rate limits.
           GH_TOKEN: ${{ inputs.alternative-gh-token-secret-name && secrets[inputs.alternative-gh-token-secret-name] || github.token }} # zizmor: ignore[overprovisioned-secrets]
       - name: C++ build
+        id: cpp-build
         run: |
           ulimit -n "$(ulimit -Hn)"
-          $INPUTS_SCRIPT
+          # shellcheck disable=SC1090
+          source "${INPUTS_SCRIPT}"
```
KyleFromNVIDIA, line 176: Please keep in mind that we are now going from `$INPUTS_SCRIPT` potentially being a space-separated command with arguments to a single file. Please make sure all existing usages are providing only a single file.

KyleFromNVIDIA (approved):
Approved with a reminder for one small pitfall

.github/workflows/conda-cpp-build.yaml
gforsyth: Yep, that's a good callout.  I did check https://github.com/search?q=org%3Arapidsai+build_cpp.sh&type=code&p=2 and there are no usages of `build_cpp.sh` with arguments passed in the `script` input.

PR #529, opened 2026-04-29
fix: add `package_name` input to cpp-post-build-check

gforsyth:
I didn't know the post-build checks bake in a call to `rapids-package-name` via `rapids-download-conda-from-github`.

`rapids-artifact-name` requires explicit arguments for the package name, so I've added that as an input. If the argument isn't provided, we maintain the current behavior.

### DIFF
--- a/.github/workflows/conda-cpp-post-build-checks.yaml
+++ b/.github/workflows/conda-cpp-post-build-checks.yaml
@@ -26,6 +26,10 @@ on:
           For example, to ignore any function symbols coming from the 'thrust' or 'cub' namespaces,
           you might provide '(void (thrust::|cub::))'.
         type: string
+      package_name:
+        description: |
+          Name of package to use when constructing artifact name using `rapids-artifact-name`, e.g. `libcudf`, `libucxx`
+        type: string
 
 defaults:
   run:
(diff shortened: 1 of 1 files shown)


gforsyth:
Tested in https://github.com/rapidsai/cudf/pull/22327/changes/66b9e07f29a22eecbb898fd5fd47cb4a901ddde6

.github/workflows/conda-cpp-post-build-checks.yaml
```diff
@@ -97,9 +101,15 @@ jobs:
...
+          PACKAGE_NAME: ${{ inputs.package_name }}
+          REPO: ${{ inputs.repo }}
           GH_TOKEN: ${{ github.token }}
         run: |
-          CPP_DIR=$(rapids-download-conda-from-github cpp)
+          if [ -n "$PACKAGE_NAME" ]; then
+            CPP_DIR=$(rapids-download-from-github "$(rapids-artifact-name conda_cpp $PACKAGE_NAME "${REPO##*/}" --cuda)")
+          else
+            CPP_DIR=$(rapids-download-conda-from-github cpp)
+          fi
```
jameslamb, line 112: Hopefully this is just temporary to allow the PR to be merged?

I'm fine with that. In general, I'd like to see there be as few different routes through the configuration as possible... e.g. I'd like the end state to be that `package_name` is required by all callers.

jameslamb (approved):

.github/workflows/conda-cpp-post-build-checks.yaml
gforsyth: Yes, this is part of the gradual roll-out -- once everything is using consistent tooling we'll remove all the conditionals.

PR #530, opened 2026-04-30
Add `build-devcontainers.yaml` and  `build-devcontainer.yaml` workflows

trxcllnt:
Add workflows to build and publish devcontainers to `ghcr.io` so we can use `cacheFrom` to speed up devcontainer launch times.

Depends on https://github.com/rapidsai/shared-actions/pull/104

### DIFF
--- a/.github/workflows/build-devcontainers.yaml
+++ b/.github/workflows/build-devcontainers.yaml
@@ -1,0 +1,96 @@
+name: Build devcontainers
+
+on:
+  workflow_call:
+    inputs:
+      cuda:
+        description: |
+          Stringified JSON array of CUDA versions to run this workflow for.
+          This is used to select .devcontainer/ directories local to wherever this workflow is invoked from.
+          For example, if a repository has directories '.devcontainer/cuda12.9-pip/' and '.devcontainer/cuda13.1-pip/',
+          '["12.9", "13.1"]' could be passed here to build both of those devcontainers in CI.
+        type: string
+        default: '["12.9", "13.1"]'
+      python_package_manager:
+        description: |
+          Stringified JSON array of Python package managers to run devcontainer builds for.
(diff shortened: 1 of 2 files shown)


.github/workflows/build-devcontainers.yaml
```diff
@@ -0,0 +1,94 @@
...
+  workflow_call:
+    inputs:
+      cuda:
+        description: |
+          Stringified JSON array of CUDA versions to run this workflow for.
+          This is used to select .devcontainer/ directories local to wherever this workflow is invoked from.
+          For example, if a repository has directories '.devcontainer/cuda12.9-pip/' and '.devcontainer/cuda13.1-pip/',
+          '["12.9", "13.1"]' could be passed here to build both of those devcontainers in CI.
+        type: string
+        default: '["12.9", "13.1"]'
```
bdice: Just a note that this may need to be changed. We're in the process of updating 13.1 to 13.2. Not all repos have the same devcontainers right now, depending on whether they've been updated.

I'd lean towards making this a required input with no default, so that we have one fewer place to update CUDA versions (both per-repo and the default value).

.github/workflows/build-devcontainers.yaml
```diff
@@ -0,0 +1,94 @@
...
+
+          # Set values to control ghcr.io cleanup below
+          cat <<EOF >> "$GITHUB_OUTPUT"
+          digests=${DIGESTS[*]}
+          name=${NAME}
+          tags=${TAGS[*]}
+          EOF
+
+      - name: Clean up untagged images
+        uses: snok/container-retention-policy@3b0972b2276b171b212f8c4efbca59ebba26eceb
```
bdice:
```suggestion
        uses: snok/container-retention-policy@3b0972b2276b171b212f8c4efbca59ebba26eceb # v3.0.1
```

bdice (approved):

PR #558, opened 2026-05-26
use full semver tags for third-party actions

jameslamb:
Proposes using full semver for tag reference comments in `uses:` blocks, to improve the usefulness of `renovate`'s output.

See #547

image

Because we weren't using the full semver tag that `docker/login-action` uses (just as an example), `renovate`'s auto-bump PR:

* shows a change in commit SHAs instead of versions
* does not auto-generate a link to release notes

## Notes for Reviewers

We're using the full version tags like this in most of the rest of RAPIDS as of #275


gforsyth (approved):
Thanks for doing this, James!

PR #563, opened 2026-06-05
Add `permissions: {packages: write}` to `build-devcontainers.yaml`

trxcllnt:
Add `permissions: {packages: write}` to `build-devcontainers.yaml`

### DIFF
--- a/.github/workflows/build-devcontainers.yaml
+++ b/.github/workflows/build-devcontainers.yaml
@@ -49,6 +49,8 @@ jobs:
     name: Clean up untagged images
     runs-on: ubuntu-latest
     continue-on-error: true
+    permissions:
+      packages: write
     steps:
       - id: vars
         name: Get image name, tags, and digests


.github/workflows/build-devcontainers.yaml
```diff
@@ -49,6 +49,8 @@ jobs:
     name: Clean up untagged images
     runs-on: ubuntu-latest
     continue-on-error: true
+    permissions:
+      packages: write
```
jameslamb, line 53: If this is merged, I think it'll require code changes in all the repos that use it too. The call site in a reusable workflow chain has to grant permissions to the workflows it calls.

Which is why you see stuff like this throughout RAPIDS:

https://github.com/rapidsai/cudf/blob/4ae9d07c30e41a77d6bc6046eeee58873ba12aac/.github/workflows/pr.yaml#L757-L762

This is described at https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#nesting-reusable-workflows

Not saying not to do this, just letting you know it'll require more than just this PR.

jameslamb (approved):
Approving so you can move this forward if you want, but see my note about all repos with devcontainers needing changes to support this.

bdice (approved):

.github/workflows/build-devcontainers.yaml
trxcllnt: I just checked the PRs in https://github.com/rapidsai/build-planning/issues/280, the repos do already pass `permissions: {packages: write}` to this workflow. That used to be enough, I'm not sure why it isn't now.

The fact that this workflow doesn't explicitly request the `packages: write` permission is the only reason I can think that this would've started failing, so this PR adds it. If it doesn't, then we may need to generate the JWT manually and require `id-token: write`.

jameslamb: Great. Looks like I was again confusing `build-in-devcontainer` (PRs) with `build-devcontainers` (`main`). I do see `packages:write` in the callsites for `build-devcontainers`, like https://github.com/rapidsai/cudf/blob/4ae9d07c30e41a77d6bc6046eeee58873ba12aac/.github/workflows/build.yaml#L464

PR #579, opened 2026-06-22
feat(renovate): add 7-day cooldown before any updates

gforsyth:
Part of rapidsai/build-planning#273

Adds a 7-day cooldown period before renovate proposes any updates

Note: there were cooldowns here already, but this declares that ALL packages should have a 7-day cooldown

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,11 +5,17 @@
   ],
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
my least favorite thing about JSON as a config format... no inline comments 😭

Issue #505, opened 2026-02-17
wheels-publish: reduce CI time


jameslamb:
## Description

The `wheels-publish` workflow is responsible for a pretty narrow purpose... downloading wheel artifacts built in CI and publishing them to package repositories (like https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/).

It really only needs a Python interpreter and a couple tools like `twine`, the `anaconda` CLI, and the `gh` CLI.

Today, it uses the `rapidsai/ci-wheel` container images ([build from rapidsai/ci-imgs](https://github.com/rapidsai/ci-imgs/blob/6f1115223530b868fe8e87d86a1cd29a02e7e9cf/ci-wheel.Dockerfile)), which are around 4.5 GB compressed (https://hub.docker.com/r/rapidsai/ci-wheel/tags).

The time to run that workflow is dominated by container-image pulling. For example, look at a recent `wheels-publish-cudf` job ([build link](https://github.com/rapidsai/cudf/actions/runs/22094821927/job/63851768615)):

<img width="676" height="652" alt="Image" src="https://github.com/user-attachments/assets/ccbf1800-56a2-4983-a620-6abb8a9b23f2" />

This issue tracks the work of designing and implementing changes to speed that up. Maybe some mix of:

* using a slimmer container image (or just `ubuntu-latest`  / `ubuntu-slim` runners without a container)
* removing telemetry
* using a shallow `git` checkout

## Benefits of this work

* reduced end-to-end time for RAPIDS nightly pipeline (https://github.com/rapidsai/workflows/actions/workflows/nightly-pipeline.yaml)
* reduced time to roll out a change across all of RAPIDS (related: https://github.com/rapidsai/build-planning/issues/238)
* reduced usage of CI resources

## Acceptance Criteria

* `wheels-publish` workflows are faster than they were on February 17, 2026

## Ideas

Just dumping some ideas here, this is not an implementation plan:

* use `ubuntu-latest` / `ubuntu-slim` GitHub-hosted runners
* use a smaller container image (like `python:latest` or `python:3.13-slim`)
* create a new `rapidsai/ci-wheel-publishing` or similar image
  - *publishing a new image does bring new legal / compliance requirements though, this might be a lot of effort*
* fiddle with the `rapidsai/ci-wheel` image
  - *different compression?*
  - *fewer layers?*
  - *removing unused things?*
* use shallow `git` checkouts
* remove telemetry (this workflow doesn't benefit much from it)

[...]

PR #585, opened 2026-06-29
remove git clones and telemetry for package-publishing jobs, use full semver tags for third-party actions

jameslamb:
Contributes to #505 

* removes telemetry and source checkouts for package-publishing jobs (`conda-upload-packages` and `wheels-publish`)

And other tiny things I noticed while doing this:

* removes outdated comments about `setuptools-scm` (RAPIDS does not use this any more)
* switches from `#v7` to `#v7.0.0` in comments on `actions/checkout` (improves clarity of auto-generated PRs to update that)
* updates all pre-commit hooks w/ `pre-commit` hooks

### Is this safe?

I think so.

`conda-upload-packages` and `wheels-publish` don't depend on anything in the calling repo's source or git history.

The `rapids-release-build` check here sort of depends on git tags:

https://github.com/rapidsai/shared-workflows/blob/ad100ac25a47afe772ed86ee69e6eabd6966a419/.github/workflows/conda-upload-packages.yaml#L110-L112

But doesn't get it from a local checkout... it reads the GitHub-specific environment variable `GITHUB_REF`: https://github.com/rapidsai/gha-tools/blob/d85a30d7a8930b079ef681a086e8f3646120c27a/tools/rapids-is-release-build#L11

### Impact

Reduces runtime of these jobs a tiny bit, and more importantly removes some network requests (a common source of CI instability).

The impact on runtime will be larger for repos where `git clone` is more expensive. For example, on a recent `cudf` build on `main`, checking out the code took 14 seconds and telemetry-related things added around 4 seconds.

image

https://github.com/rapidsai/cudf/actions/runs/28378826926/job/84082197842

### How I tested this

Proposing testing this here: https://github.com/rapidsai/rmm/pull/2458

### DIFF
--- a/.github/workflows/wheels-publish.yaml
+++ b/.github/workflows/wheels-publish.yaml
@@ -84,11 +84,14 @@ jobs:
         RAPIDS_BUILD_TYPE: ${{ inputs.build_type }}
     steps:
     - name: checkout code repo
-      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       with:
         repository: ${{ inputs.repo }}
         ref: ${{ inputs.sha }}
-        fetch-depth: 0 # unshallow fetch for setuptools-scm
+        fetch-depth: 1
+        # fetch all tags even though we're only fetching 1 commit
+        # (for use in logic that checks recent tags to determine versions)
+        fetch-tags: true
         persist-credentials: false
 
     - name: Telemetry setup
--- a/.github/workflows/wheels-publish.yaml
+++ b/.github/workflows/wheels-publish.yaml
@@ -91,7 +91,9 @@ jobs:
         branch: ${{ inputs.branch }}
         build_workflow_name: ${{ inputs.build_workflow_name }}
         date: ${{ inputs.date }}
-        sha: ${{ inputs.sha }}
+        # On 'push' or 'tag', 'inputs.sha' is null.
+        # Fall back to 'github.sha' to avoid needing a repo checkout.
+        sha: ${{ inputs.sha || github.sha }}
 
     # Per the docs at https://docs.github.com/en/rest/rate-limit/rate-limit?apiVersion=2022-11-28#get-rate-limit-status-for-the-authenticated-user,
     # checking '/rate_limit | jq .' should not itself count against any rate limits.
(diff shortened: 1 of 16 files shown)


gforsyth (approved):
I'm all for removing the unnecessary checkouts and adding the full tags for the rest.

As for removing `telemetry`, I think I'm a +0 -- I haven't seen that particular step fail very often, but it's a job that only runs on `main` and `release/` branches. 

If we do have those upload jobs instrumented, I'm trying to think of what we might catch.

If they start taking longer, could mean:
1. Artifacts are larger, but we'd catch that anyway from `pydistcheck`
2. Network is having issues (but we'd see that everywhere)

Yeah, I'm good with this.  Thanks, James!

.github/workflows/wheels-publish.yaml
```diff
@@ -105,7 +91,9 @@ jobs:
         branch: ${{ inputs.branch }}
         build_workflow_name: ${{ inputs.build_workflow_name }}
         date: ${{ inputs.date }}
-        sha: ${{ inputs.sha }}
+        # On 'push' or 'tag', 'inputs.sha' is null.
+        # Fall back to 'github.sha' to avoid needing a repo checkout.
+        sha: ${{ inputs.sha || github.sha }}
```
jameslamb, line 96: The first build of that test RMM PR failed like this:

```text
RAPIDS_REPOSITORY=rapidsai/rmm
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Error: Process completed with exit code 128.
```

([build link](https://github.com/rapidsai/rmm/actions/runs/28388699537/job/84110880953))

If the `sha` input to `rapidsai/shared-actions/rapids-github-info` is null or an empty string, it falls back to running `git rev-parse HEAD`.

https://github.com/rapidsai/shared-actions/blob/113b75588021524a58ebbd8c4b91fb46eb464a0c/rapids-github-info/action.yml#L34-L36

It'll be null when this workflow is triggered by `push` (merge to `main`) or `tag`.

In those cases, I think it's safe and appropriate to fall back to `github.sha` if `inputs.sha` is not provided. Based on my read of https://docs.github.com/en/actions/reference/workflows-and-actions/contexts, that should be the value we want in `push` or `tag` contexts.

I restarted `rmm` CI, let's see what it says: https://github.com/rapidsai/rmm/actions/runs/28388699537

jameslamb: That worked!

And looks to me like the correct SHA and versions were chosen

Image

https://github.com/rapidsai/rmm/commits/main/

conda:

Image

wheels:

Image

https://github.com/rapidsai/rmm/actions/runs/28388699537

gforsyth (approved):
Agree with your additional analysis and think this is good to go!

PR #593, opened 2026-07-15
ci(conda-python-tests): add script-env-secret-4 slot

ramakrishnap-nv:
Hit a scenario where 3 secret slots were already taken by other credentials and a 4th was needed. Adds `script-env-secret-4-key`/`script-env-secret-4-value` following the existing pattern.

### DIFF
--- a/.github/workflows/conda-python-tests.yaml
+++ b/.github/workflows/conda-python-tests.yaml
@@ -100,6 +100,16 @@ on:
           Secret value.
           Will be available via an environment variable named by 'secrets.script-env-secret-3-key'.
         required: false
+      script-env-secret-4-key:
+        description: |
+          Name of an environment variable in the environment where 'inputs.script' is run.
+          Variable's value will be set to the value passed as 'secrets.script-env-secret-4-value'.
+        required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/conda-python-tests.yaml
```diff
@@ -250,8 +261,6 @@ jobs:
         if: always()
       - name: Run codecov
         if: inputs.run_codecov
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
KyleFromNVIDIA, line 254: Before merging, we need to make sure no other projects are relying on the `CODECOV_TOKEN` secret.

ramakrishnap-nv: Good point. Updated to keep the line but use `${{ env.CODECOV_TOKEN || secrets.CODECOV_TOKEN }}` — existing callers using `secrets: inherit` fall back to `secrets.CODECOV_TOKEN` as before, while callers using `script-env-secret-4` get the value from `GITHUB_ENV`.

KyleFromNVIDIA (approved):

PR #597, opened 2026-07-22
X-ORG-596: publish-api-docs workflow

josephine-wolf-oberholtzer:
Closes #596

### DIFF
--- a/.github/workflows/publish-api-docs.yaml
+++ b/.github/workflows/publish-api-docs.yaml
@@ -1,0 +1,182 @@
+name: Publish API docs
+
+on:
+  workflow_call:
+    inputs:
+      dry-run:
+        default: "false"
+        description: If true, run without making any changes
+        required: false
+        type: boolean
+      source-path:
+        default: docs
+        description: Local path to download docs to
+        required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-api-docs.yaml
```diff
@@ -0,0 +1,183 @@
...
+        default: rapidsai-docs
+        description: The S3 bucket to download files from
+        required: false
+        type: string
+      source-s3-key:
+        description: The S3 key to download files from, e.g. the repo name
+        required: true
+        type: string
+      version-map:
+        default: '{"26.08": "legacy", "26.10": "stable"}'
```
bdice: I'd like to have no defaults here, and to force the caller to provide (and update) these values.

At the very least, we must have all hardcoded versions in this repository automatically update when https://github.com/rapidsai/shared-workflows/blob/main/ci/release/update-version.sh is called.

josephine-wolf-oberholtzer: You got it. I'll move the map to an example in the input description.

bdice (approved):

PR #600, opened 2026-07-22
X-ORG-596: finesse publish-api-docs workflow

josephine-wolf-oberholtzer:
Contributes to #596 

- The source S3 path on the RAPIDSAI side needs to include `html/` in the path
- Let's make it simpler to publish multiple docs sets by converting the entire workflow to a matrix execution
- Let's run the publish job inside a CI container to make manually installing gha-tools unnecessary
- Let's publish numeric, tagged, "latest", and "nightly" versions, depending on the branch
    - always publish numeric version
    - if numeric version aligns with a "tag" (e.g. "stable"), publish to that tag
    - if the tag is "stable", publish to "latest"
    - if the branch is "main", publish to "nightly"
    - (all of the above gated by whether dry-run is true/false)

### DIFF
--- a/.github/workflows/publish-api-docs.yaml
+++ b/.github/workflows/publish-api-docs.yaml
@@ -145,7 +145,7 @@ jobs:
 
     # publish <project>/<version>
     - name: Publish numeric version to docs.nvidia.com
-      uses: rapidsai/shared-actions/publish-docs@josephine/x-org-123-finesse-shared-action-more
+      uses: rapidsai/shared-actions/publish-docs@main
       with:
         akamai-access-token: ${{ secrets.akamai-access-token }}
         akamai-client-secret: ${{ secrets.akamai-client-secret }}
@@ -165,7 +165,7 @@ jobs:
     # publish <project>/<tag> (e.g. stable, legacy)
     - name: Publish tagged version to docs.nvidia.com
       if: steps.version-tag.outputs.rapids-version-tag != '' && startsWith(github.ref_name, 'release/')
-      uses: rapidsai/shared-actions/publish-docs@josephine/x-org-123-finesse-shared-action-more
+      uses: rapidsai/shared-actions/publish-docs@main
       with:
         akamai-access-token: ${{ secrets.akamai-access-token }}
         akamai-client-secret: ${{ secrets.akamai-client-secret }}
@@ -185,7 +185,7 @@ jobs:
     # publish <project>/latest
     - name: Publish "latest" version to docs.nvidia.com
       if: steps.version-tag.outputs.rapids-version-tag == 'stable'
-      uses: rapidsai/shared-actions/publish-docs@josephine/x-org-123-finesse-shared-action-more
+      uses: rapidsai/shared-actions/publish-docs@main
       with:
         akamai-access-token: ${{ secrets.akamai-access-token }}
         akamai-client-secret: ${{ secrets.akamai-client-secret }}
@@ -205,7 +205,7 @@ jobs:
     # publish <project>/nightly
     - name: Publish "nightly" version to docs.nvidia.com
       if: github.ref_name == 'main'
-      uses: rapidsai/shared-actions/publish-docs@josephine/x-org-123-finesse-shared-action-more
+      uses: rapidsai/shared-actions/publish-docs@main
       with:
         akamai-access-token: ${{ secrets.akamai-access-token }}
         akamai-client-secret: ${{ secrets.akamai-client-secret }}
--- a/.github/workflows/publish-api-docs.yaml
+++ b/.github/workflows/publish-api-docs.yaml
@@ -182,9 +182,9 @@ jobs:
         target-s3-key-suffix: ${{ steps.version-tag.outputs.rapids-version-tag }}
         target-s3-key: ${{ matrix.project }}
 
-    # publish <project>/latest if stable
+    # publish <project>/latest if main, same as nightly
     - name: Publish "latest" version to docs.nvidia.com
-      if: steps.version-tag.outputs.rapids-version-tag == 'stable' && startsWith(github.ref_name, 'release/')
+      if: github.ref_name == 'main'
       uses: rapidsai/shared-actions/publish-docs@main
       with:
         akamai-access-token: ${{ secrets.akamai-access-token }}
--- a/.github/workflows/publish-api-docs.yaml
+++ b/.github/workflows/publish-api-docs.yaml
@@ -162,7 +162,7 @@ jobs:
         target-s3-key-suffix: ${{ steps.version.outputs.rapids-version-major-minor }}
         target-s3-key: ${{ matrix.project }}
 
-    # publish <project>/<tag> (e.g. stable, legacy)
+    # publish <project>/<tag> (e.g. stable, legacy) if in version map
     - name: Publish tagged version to docs.nvidia.com
       if: steps.version-tag.outputs.rapids-version-tag != '' && startsWith(github.ref_name, 'release/')
       uses: rapidsai/shared-actions/publish-docs@main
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-api-docs.yaml
```diff
@@ -51,9 +61,6 @@ on:
       target-aws-region:
         description: AWS region
         required: true
-      target-aws-role-to-assume:
-        description: AWS role to assume
-        required: true
```
rockhowse, line 56: interesting removal here... so in this new method we are just using straight up AWS access keys vs the more modern assume role pattern (temporary auth + short lived token?)

.github/workflows/publish-api-docs.yaml
```diff
@@ -66,24 +73,40 @@ defaults:
...
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
+      - name: Set matrix
+        id: set-matrix
+        env:
+          DOCS_PROJECTS: ${{ inputs.docs-projects }}
+        run: |
+          matrix=$(echo "${DOCS_PROJECTS}" | jq -Rc 'split(",")')
+          echo "matrix=${matrix}" >> "${GITHUB_OUTPUT}"
```
rockhowse, line 87: Is one of the major changes here so that instead of having one giant docs build, we are splitting out each repos' docs into their own parallel matrix run?

.github/workflows/publish-api-docs.yaml
```diff
@@ -114,70 +137,87 @@ jobs:
     - name: Download docs
       env:
         S3_BUCKET: ${{ inputs.source-s3-bucket }}
-        S3_KEY: ${{ inputs.source-s3-key }}
+        S3_KEY: ${{ matrix.project }}
```
rockhowse, line 140: in this context is `project` equivliant to a `repo` or is it more aligned with higher level groupings `cuvs` + `cuvs-lucene` + `raft` == single project `cuvs` or would this key map to one entry per repo?

.github/workflows/publish-api-docs.yaml
```diff
@@ -114,70 +137,87 @@ jobs:
...
+        akamai-host: ${{ secrets.akamai-host }}
+        akamai-request-name: ${{ steps.akamai.outputs.request-name }}-nightly
+        dry-run: ${{ inputs.dry-run }}
+        source-path: ${{ inputs.source-path }}
+        target-aws-access-key-id: ${{ secrets.target-aws-access-key-id }}
+        target-aws-region: ${{ secrets.target-aws-region }}
+        target-aws-secret-access-key: ${{ secrets.target-aws-secret-access-key }}
+        target-s3-bucket: ${{ secrets.target-s3-bucket }}
+        target-s3-key-suffix: nightly
+        target-s3-key: ${{ matrix.project }}
```
rockhowse, line 223: The diff on this is a little complex to trace through, but let me just see if understand the intent here

Original behavior
* `nightly` - if we were on the `main` branch
* `tagged`- if the ref started with `release/`

New behavior looks to add one more doc publishing layer `latest` if the tag is `stable` and the ordering of the blocks changed slightly
* `tagged`- if the ref started with `release/`
* `latest` - if the tag is `stable` (new) 
* `nightly` - if we were on the `main` branch

Effectively preserving the original two publications with a new third one.

rockhowse (commented):
I don't think I should be the authoritative ✅ on this but read through it and here's the summary of what this looks to accomplish 

* migrate from a single one-shot docs publish to a matrix of parallel publications (ideally decreases publication time) 
* change the auth mechanism to remove assume role with short lived tokens and instead use a longer lived token (probably pertaining to us being in an Nvidia wide AWS account instead of one owned by `RAPIDS`) 
* added a new `stable/latest` publication to the existing `nightly` and `tagged/release` publications

.github/workflows/publish-api-docs.yaml
josephine-wolf-oberholtzer: Yes, this is not my preferred setup. I would rather do it the way you describe, and the way other teams pushing docs to S3 do it, but my hands are tied for the moment.

.github/workflows/publish-api-docs.yaml
josephine-wolf-oberholtzer: Some RAPIDS repos publish multiple different sets of docs. For example, https://github.com/rapidsai/rmm publishes https://docs.rapids.ai/api/rmm/stable/ and https://docs.rapids.ai/api/librmm/stable/

Rather than having one workflow call per docs "target", I'm simplifying things by letting `RMM` say "I publish rmm and librmm, handle it".

.github/workflows/publish-api-docs.yaml
josephine-wolf-oberholtzer: Lower level than a repo. See: https://github.com/rapidsai/shared-workflows/pull/600#discussion_r3667744570

.github/workflows/publish-api-docs.yaml
josephine-wolf-oberholtzer: Yes, that's the gist of it. I'm adding `latest` here because that gives us a pivot toward docs that are tagged more similarly to the rest of the docs.nvidia.com ecosystem.

.github/workflows/publish-api-docs.yaml
rockhowse: ahh got it, this is a "shared workflow" and NOT a central pipeline that publishes all docs. So repos probably have their own per/repo set of invocations. I like ti!

.github/workflows/publish-api-docs.yaml
rockhowse: totall understand, just calling out a major change in access in case it was done unexpectedly. Since it's purposeful and I know we have many constraints on all quarters not a blocker by any means.

.github/workflows/publish-api-docs.yaml
rockhowse: ok makes more sense thx for the clarification 👍🏼

.github/workflows/publish-api-docs.yaml
rockhowse: LOL yeah I see the "latest" vs "historic" pattern with things like CUDA-toolkit releases all and it's been a point of frustration over the years... I wish for some of these docs they had both the `latest` AND the latest version linked by the version number itself available. Because of this, you have to wait for the NEXT release to get the deep linking to the `latest` docs themselves (T-1 release) (many of my various "configuring GDS" docs have broken because of this 😢 ) Glad we have BOTH supported and having the `latest` variant makes total sense to me having been a customer of Nvidia for years now 😄

.github/workflows/publish-api-docs.yaml
```diff
@@ -114,70 +137,87 @@ jobs:
...
         target-s3-bucket: ${{ secrets.target-s3-bucket }}
-        target-s3-key-suffix: nightly
-        target-s3-key: ${{ inputs.source-s3-key }}
+        target-s3-key-suffix: ${{ steps.version-tag.outputs.rapids-version-tag }}
+        target-s3-key: ${{ matrix.project }}
 
-    # publish <project>/<tag> (e.g. stable, legacy)
-    - name: Publish tagged version to docs.nvidia.com
-      if: steps.version-tag.outputs.rapids-version-tag != '' && startsWith(github.ref_name, 'release/')
+    # publish <project>/latest if stable
```
bdice: Eh? I know our naming system is a little inconsistent but I was thinking latest=nightly and not latest=stable. I might be missing context here.

I think we have three names in RAPIDS. Nightly, stable, and legacy (which we're not focused on yet). I think "latest" is ambiguous. (Is it latest release or latest `main`?)

Also note that when we use the word "latest" in our container labels like `26.10-latest`, we actually mean "latest dependencies." We are saying "this is our RAPIDS 26.10 image with the latest CUDA, latest Python, latest OS, etc."

bdice (approved):
Code generally looks fine up to the naming question. Approving to allow continued iteration as needed.

.github/workflows/publish-api-docs.yaml
josephine-wolf-oberholtzer: Sounds good. I see different (and generally inconsistent) behavior on other docs on docs.nvidia.com, but happy to swap.

PR #602, opened 2026-07-23
feat(conda): add `dual_publish` input to publish release builds to release and nightly

gforsyth:
We want to publish some of our packages to both `rapidsai` and `rapidsai-nightly` conda channels, namely `rapids-logger` and `rapids-build-backend` (possibly more).

Currently we check if it's a release build, then dispatch to one or the other, accordingly (the channel destination is determined by the token used).

This adds an additional upload step -- both are conditional on the output of `rapids-is-release-build`, but now there is a `dual_publish` input that will allow also pushing the release builds to the nigthly channel.


KyleFromNVIDIA (approved):

jameslamb (approved):
love this let's do it

PR #603, opened 2026-07-27
builds: add support for computing a date-based version component at build time

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/218

Proposes adding a new workflow, `compute-build-details`, which computes information to be passed into builds.

Right now this only calculates a version suffix like `.post{current-time-in-epoch-seconds}` to help with distinguishing nightlies, but proposing giving it a generic workflow name in case we want to stick other data in here in the future.

### How does this solve the problem with nightlies?

With this time-based version component, every new non-release run of `build.yaml` in a RAPIDS repo should produce packages with a never-before-seen version, so uploads to anaconda.org should succeed.

e.g. we might see this version sequence for nightlies:

```text
26.08.00a35.post1785176675  # 35 commits past '26.08.00a' tag, CI run started at epoch time 1785176675
26.08.00a35.post1785180275  # 35 commits past '26.08.00a' tag, CI run started 1 hour later
26.08.00a36.post1785180475  # 36 commits past '26.08.00a' tag, CI run started a little later
```

These versions will only be used in pre-releases built on `main`, not releases or packages built in PR CI (https://github.com/rapidsai/gha-tools/pull/268).

### What about timing problems?

This workflow will run once at the very beginning of the branch/nightly workflow run, then its output is passed as input to all other build jobs.

That makes the value static across an entire workflow run, so packages will have the same version even for:

* builds starting at different times due to scheduling delays
* builds starting at different times from clicking "re-run failed jobs" (e.g. for network failures)

### How I tested this

Confirmed that versions like this are recognized as valid PEP 440 versions and sort the way we expect in `packaging.Version`.

```python
from packaging.version import Version

Version("2026.08.00") > Version("2026.08.00a35.post1785176675")
# True

Version("2026.08.00rc0") > Version("2026.08.00a35.post1785176675")
# True

# same commit, newer build wins
Version("2026.08.00a35.post1885176675") > Version("2026.08.00a35.post1785176675")
# True

# later commit wins
Version("2026.08.00a36.post1885176675") > Version("2026.08.00a35.post1785176675")
# True
```

Also tested live with on `cugraph-gnn`, see "How I tested this" on https://github.com/rapidsai/cugraph-gnn/pull/508

### DIFF
--- a/.github/workflows/compute-build-details.yaml
+++ b/.github/workflows/compute-build-details.yaml
@@ -16,4 +16,4 @@ jobs:
       - name: build
         id: build-details
         run: |
-          echo "version-suffix=.post$(date +%s)" | tee -a "${GITHUB_OUTPUT}"
+          echo "version-suffix=.post$(date -u +'%y%m%d%H%M%S')" | tee -a "${GITHUB_OUTPUT}"
--- a/.github/workflows/compute-build-details.yaml
+++ b/.github/workflows/compute-build-details.yaml
@@ -1,0 +1,19 @@
+on:
+  workflow_call:
+    outputs:
+      package-version-suffix:
+        description: "Suffix to be appended to conda/sdist/wheel version numbers."
+        value: ${{ jobs.build.outputs.version-suffix }}
+
+permissions: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-slim
+    outputs:
+      version-suffix: ${{ steps.build-details.outputs.version-suffix }}
+    steps:
+      - name: build
+        id: build-details
+        run: |
+          echo "version-suffix=.post$(date +%s)" | tee -a "${GITHUB_OUTPUT}"
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -16,6 +16,14 @@ on:
       sha:
         description: "Full git commit SHA to check out"
         type: string
+      version-suffix:
+        description: |
+          Optional value to be appended to package versions.
+          Will be appended exactly, so e.g. to add a '.' separator,
+          provide '.post1' not 'post1'.
+        default: null
+        required: false
+        type: string
       repo:
         description: "Git repo to check out, in '{org}/{repo}' form, e.g. 'rapidsai/cudf'"
         type: string
@@ -101,6 +109,7 @@ jobs:
       image: rapidsai/ci-conda:26.10-cuda${{ matrix.CUDA_VER }}-${{ matrix.LINUX_VER }}-py${{ matrix.PY_VER }}
       env:
         RAPIDS_BUILD_TYPE: ${{ inputs.build_type }}
+        RAPIDS_VERSION_SUFFIX: ${{ inputs.version-suffix }}
     steps:
       - uses: aws-actions/configure-aws-credentials@517a711dbcd0e402f90c77e7e2f81e849156e31d # v6.2.2
         with:
--- a/.github/workflows/conda-python-build.yaml
+++ b/.github/workflows/conda-python-build.yaml
@@ -16,6 +16,14 @@ on:
       sha:
         description: "Full git commit SHA to check out"
         type: string
+      version-suffix:
+        description: |
+          Optional value to be appended to package versions.
+          Will be appended exactly, so e.g. to add a '.' separator,
+          provide '.post1' not 'post1'.
+        default: null
+        required: false
+        type: string
       repo:
         description: "Git repo to check out, in '{org}/{repo}' form, e.g. 'rapidsai/cudf'"
         type: string
@@ -111,6 +119,7 @@ jobs:
       image: rapidsai/ci-conda:26.10-cuda${{ matrix.CUDA_VER }}-${{ matrix.LINUX_VER }}-py${{ matrix.PY_VER }}
       env:
         RAPIDS_BUILD_TYPE: ${{ inputs.build_type }}
+        RAPIDS_VERSION_SUFFIX: ${{ inputs.version-suffix }}
     steps:
       - uses: aws-actions/configure-aws-credentials@517a711dbcd0e402f90c77e7e2f81e849156e31d # v6.2.2
         with:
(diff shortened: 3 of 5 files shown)


.github/workflows/compute-build-details.yaml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-slim
+    outputs:
+      version-suffix: ${{ steps.build-details.outputs.version-suffix }}
+    steps:
+      - name: build
+        id: build-details
+        run: |
+          echo "version-suffix=.post$(date +%s)" | tee -a "${GITHUB_OUTPUT}"
```
jameslamb: Thread to talk about options I considered for versioning.

```text
# epoch seconds
.post1785251654

# epoch milliseconds
.post17852516540000

# integer-ified human-readable datetime (American ordering of month / day 😛, minute precision)
.post20260728142801

# integer-ified human-readable datetime like above, 2-digit year
.post260728142801

# etc., etc. (variations on precision and 2-vs.-4 digit year
```

Proposing epoch seconds because:

* it's more than enough precision (we can accept the risk of 2 builds being triggered within the same second and having colliding versions)
* it won't overflow max `int32` until 2038 ([link](https://en.wikipedia.org/wiki/Year_2038_problem)), giving us some protection in the case that some tool that processes versions is representing components as `int32` values
* it leaves open the possibility of switching to any of the other forms in the future with 0 impact on correctness (all the other forms are strictly larger)

But I don't feel that strongly about this. Happy to hear arguments in favor of a different format.

The properties we want are:

* valid PEP 440 identifier, parseable by every tool that touches these versions
* packages built later have versions that lead `conda` and `pip` solvers to prefer them to earlier packages
* packages built from a further-along commit are preferred to those built from earlier commits

bdice: To compare prior art, we use `YYMMDD` in `rapids-date-string`, which we use to generate build strings in our conda packages. https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-date-string

I think the proposal here is reasonable, though I could see an argument for something like `YYMMDDhhmmss` that we could use in both conda and wheel packages. It'd make it much cleaner to align the builds and know when each was produced.

bdice: The only "what if" situation I am concerned about is "rerun all jobs" instead of "rerun failed jobs." Does that reset the clock? Rather than calling `date`, should we fetch some property of the GitHub Actions job (if not running locally) that corresponds to the run (but not the run attempt)? We would still need to fall back to `date` for local runs.

I looked at [a job that had reruns](https://github.com/rapidsai/cudf/actions/runs/30341971248). It seems like `gh api repos/rapidsai/cudf/actions/runs/30341971248 --jq '.created_at'` is fixed at the first run, so it is equivalent to `gh api repos/rapidsai/cudf/actions/runs/30341971248/attempts/1 --jq '.run_started_at'`. The second rerun has the same `created_at` value but a different `run_started_at` value.

jameslamb: > _I could see an argument for something like YYMMDDhhmmss that we could use in both conda and wheel packages_

Yeah I was going back and forth on that. `YYMMDDhhmmss` is bigger than max int32, that's the only thing that gave me pause. I was worried about some tool somewhere in our stack using an int32 for version components and overflowing or raising some exception when processing such a large number.

I'll try it here, I have a good test setup going in that `cugraph-gnn` PR that'll touch `pip`, `rattler`, and `conda`, that should be enough for us to build some confidence.

> _rerun all jobs_

What's wrong with resetting the clock?

From my perspective, that's not a problem. As long as all the packages produced from a run have the same version and that is newer than any previously-published nightlies, I don't see think we need to worry about that version being "later" than the one that would have been produced if CI had succeeded without needing "rerun all jobs".

It looks to me like `created_at` and `started_at` are not available as context (https://docs.github.com/en/actions/reference/workflows-and-actions/contexts), so those would actually require hitting the GitHub API like in your example.

Unless there's a really compelling reason, I'd prefer not to introduce more GitHub API calls and all the costs those introduce here (e.g. configuring retries, auth token, contribution towards rate limits, etc.).

KyleFromNVIDIA: IMO, two runs of the same workflow should produce different timestamps. If some jobs in a workflow succeed, and then we re-run it, aren't we then back to trying to overwrite existing packages since the timestamp is the same?

KyleFromNVIDIA: I think it's good the way it currently stands. It doesn't require hitting the GitHub API, keeps the same date within a workflow run (even if it was rerun, as long as it wasn't rerun-all), and generates a new date if we rerun all.

KyleFromNVIDIA: The one caveat is that I would also support a `YYMMDDhhmmss` format.

bdice: > worried about some tool somewhere in our stack using an int32 for version components

I can't think of any places where that would come up. I would lean towards human-readable dates first and then fall back to epochs if needed.

> What's wrong with resetting the clock?

I think the points you and Kyle raised are totally valid. I had some hesitation about breaking the 1-1 relationship between artifacts and workflows, but maybe it's fine for independent runs to generate new packages. I would treat `created_at` as something for us to keep at the back of our minds if we run into problems.

jameslamb: Alright thanks all, just pushed https://github.com/rapidsai/shared-workflows/pull/603/commits/68ffdd3e1f9460687acbb370100d30cc710075a2 changing these to a `YYMMDDhhmmss` format.

I'll go test that in the `cugraph-gnn` PR.

vyasr: I agree with this overall approach and using the `YYMMDDhhmmss` as a good format.

vyasr (approved):

.github/workflows/compute-build-details.yaml
jameslamb: > _I'll go test that in the `cugraph-gnn` PR_

Testing on https://github.com/rapidsai/cugraph-gnn/pull/508 went well, did not see any issues with these versions there. That covers most of what's typical in RAPIDS CI, so I think that's a good sign that this could be safe to roll out.

PR #604, opened 2026-07-29
enforce 'yamllint' checks

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/305

Proposes enforcing `yamllint` checks here, to standardize formatting of YAML files and catch correctness issues.


bdice (approved):
#600 will have conflicts here, please work with @josephine-wolf-oberholtzer on merge timing.

PR #605, opened 2026-07-29
conda-upload-packages: fix nightly uploads

jameslamb:
Uploads of nightly conda packages stopped working after https://github.com/rapidsai/shared-workflows/pull/602 was merged (only affects `main`).

Thankfully we have integration tests to catch exactly this type of thing, and they caught it: https://github.com/rapidsai/integration/issues/861

This fixes that.

### DIFF
--- a/.github/workflows/conda-upload-packages.yaml
+++ b/.github/workflows/conda-upload-packages.yaml
@@ -140,7 +140,7 @@ jobs:
             echo "is_release=false" >> "${GITHUB_OUTPUT}"
           fi
       - name: Upload packages to release location
-        if: ${{ steps.conda-upload-destination.outputs.is_release == true }}
+        if: ${{ steps.conda-upload-destination.outputs.is_release == 'true' }}
         run: rapids-upload-to-anaconda-github
         env:
           SKIP_UPLOAD_PKGS: ${{ inputs.skip_upload_pkgs }}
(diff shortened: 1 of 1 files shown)


.github/workflows/conda-upload-packages.yaml
```diff
@@ -140,15 +140,15 @@ jobs:
...
+        if: ${{ steps.conda-upload-destination.outputs.is_release == 'true' }}
         run: rapids-upload-to-anaconda-github
         env:
           SKIP_UPLOAD_PKGS: ${{ inputs.skip_upload_pkgs }}
           RAPIDS_CONDA_UPLOAD_LABEL: ${{ inputs.upload_to_label }}
           RAPIDS_CONDA_TOKEN: ${{ secrets.CONDA_RAPIDSAI_TOKEN }}
           GH_TOKEN: ${{ github.token }}
       - name: Upload packages to nightly location
-        if: ${{ steps.conda-upload-destination.outputs.is_release == false || inputs.dual_publish }}
+        if: ${{ steps.conda-upload-destination.outputs.is_release == 'false' || inputs.dual_publish }}
```
jameslamb, line 151: GitHub Actions step outputs become strings, so you can't do boolean comparison like `== false`. 

See https://docs.github.com/en/actions/reference/workflows-and-actions/expressions, which says:

> _`steps.<step_id>.outputs.<output_name>` evaluates as a string. You need to use specific syntax to tell GitHub to evaluate an expression rather than treat it as a string. For more information, see [Contexts reference](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#steps-context)._

Sorry for missing it in review.

bdice: Is there any automation that can check for things like this?

bdice (approved):

gforsyth (approved):
Thanks for catching and fixing this, James!

:shipit:

.github/workflows/conda-upload-packages.yaml
jameslamb: I haven't been able to find one.

Looks like it's been mentioned in `actionlint` before (https://github.com/rhysd/actionlint/issues/147) but wasn't added there because it's technically allowed even though it could be surprising.

PR #607, opened 2026-07-29
Add a shared workflow for publishing to Maven Central

msarahan:
This PR replaces #606 with a branch on the core repo so that CI can run.

* Add `.github/workflows/maven-publish.yaml`, a reusable workflow that
  publishes release-candidate Maven artifacts (jars, POMs, sources,
  javadoc) to Maven Central via the Sonatype Central Publisher Portal.
  Callers invoke it on tag builds.

* Two-phase pipeline: sign and upload to Artifactory first, then
  forward the same bytes to the Publisher Portal. Artifactory acts as
  the audit layer. Every upload lands at
  `<groupPath>/<artifactId>/<version>/`, overwriting any earlier
  bundle at the same coordinate (the workflow logs when it does so)
  so a later RC always wins. Each file is stamped with a
  `source.git-sha` matrix property linking it back to a commit.

* Upload runs in a container because it needs a Maven client and GPG.
  Promote runs directly on the host runner because it needs only an
  HTTP client. This avoids paying container startup cost twice.

* The workflow never calls the Portal's `/publish` endpoint. The
  `stage-for-maven-central-publish` input picks between two modes: by
  default it validates the bundle and immediately drops it (a safe
  rehearsal), or it can leave the bundle PENDING for a human to
  release from the Portal UI. Either way, promoting to Central
  requires a human click.

* Generate `.md5` and `.sha1` sidecars locally at promote time.
  Sonatype requires them for every file (including `.asc` signatures),
  and Artifactory doesn't preserve them as separately-listable files,
  so the pipeline generates them on the bundle right before
  uploading.

* The workflow resolves its own repository and ref from the caller's
  `uses:` line rather than hardcoding `rapidsai/shared-workflows`, so
  callers testing from a fork automatically pick up their own copy of
  the pipeline scripts.

* Add `RAPIDS_ARTIFACTORY_URL` and `RAPIDS_ARTIFACTORY_REPO` to the
  `config-variables` list in `.github/actionlint.yaml` so actionlint
  recognizes the workflow's references to them.

Closes https://github.com/rapidsai/build-infra/issues/382, https://github.com/rapidsai/build-infra/issues/383


paul-aiyedun (approved):
The last workflow run succeeded (https://github.com/rapidsai/cudf/actions/runs/30669539430/job/91284511980?pr=23447), and the bundle was verified to be staged for deployment on Maven Central.

PR #609, opened 2026-08-03
release: generate release catalog companions for build artifacts

msarahan:
**Posted by Codex (GPT-5) on behalf of [@msarahan](https://github.com/msarahan). This pull request description is LLM-generated; readers should treat its content accordingly.**

## Why

Candidate-first releases need an immutable record of the exact bytes produced by CI, rather than requiring a release coordinator to reconstruct a package list and download artifacts later. This adds a release catalog companion beside each existing build artifact so the release platform can collect the already-built packages, their source revision, and available evidence from the exact workflow run.

## What changes

- Conda C++, Conda Python, and wheel producers upload a `release-catalog-<source-artifact-name>` companion whenever they upload their normal source artifact.
- Standard producers set `release_catalog_key` to `conda:<repository-name>` or `wheel:<repository-name>`, pass their producer output as `artifact_directory`, and let the action discover and parse every built package independently.
- `custom-job.yaml` remains opt-in through the `release-catalog-config` reusable-workflow input. An empty value disables catalog generation; a producer passes one JSON object when its uploaded bundle is a publishable release artifact set.
- Explicit artifact descriptors carry their own identity and evidence references. Parseable Conda and wheel artifacts derive identity from the package itself; other artifacts provide `package_identity_file` on that artifact's descriptor.
- Each companion contains one enveloped `release-catalog-entries.json` and available provenance, signature, and SBOM sidecars without changing the producer's existing artifact bundle.

Configuration is validated against the [`shared-actions` release catalog schema](https://github.com/rapidsai/shared-actions/blob/ed0c8f1fdeeff3348e50ea8fe5032c74b18542a7/release-catalog/config.schema.json) before artifacts are inspected.

The workflows pass `source-sha: ${{ env.RAPIDS_SHA }}` to the action. `RAPIDS_SHA` is populated from the caller-provided checkout SHA when available and otherwise falls back to `git rev-parse HEAD`, so the catalog records the revision actually checked out by the build.

Standard Conda and wheel callers need no release-specific artifact selection. Multiple packages built by the same repository share its repository-level catalog key unless they have independently managed versioning, validation, dependency ordering, publication destination, or promotion strategy.

Tracks [`rapidsai/build-infra#381`](https://github.com/rapidsai/build-infra/issues/381).

### DIFF
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -48,7 +48,10 @@ on:
         type: string
         default: ''
         required: false
-        description: "Override the release-platform unit ID; defaults to conda:<repository-name>"
+        description: >-
+          Release component ID: a string label, not a file or bundle. It is written to each manifest entry and groups
+          files and matrix variants for release assembly. Leave empty to use conda:<repository-name>; override only
+          when one repository's Conda producers must be tracked as distinct components.
       matrix_filter:
         description: |
           jq expression which modifies the matrix.
@@ -234,11 +237,14 @@ jobs:
 
       - name: Create Conda release build-output companion
         if: ${{ inputs.upload-artifacts }}
-        uses: rapidsai/shared-actions/release-build-output-dispatch@3a9568fc2c9bc9fd05e4cf76a73d9ad481f93ac5 # agent/release-build-output-container
+        uses: rapidsai/shared-actions/release-build-output-dispatch@a18a4a7ac572366c09c15641ec274cc6f15bfb5d # agent/release-build-output-container
         with:
-          artifact-type: conda
-          output-directory: ${{ steps.package-name.outputs.CONDA_OUTPUT_DIR }}
-          release-unit: ${{ inputs.release-unit || format('conda:{0}', github.event.repository.name) }}
+          config: >-
+            {
+              "artifact_type": "conda",
+              "component_id": ${{ toJSON(inputs.release-unit || format('conda:{0}', github.event.repository.name)) }},
+              "output_directory": ${{ toJSON(steps.package-name.outputs.CONDA_OUTPUT_DIR) }}
+            }
           source-artifact-name: ${{ steps.package-name.outputs.RAPIDS_PACKAGE_NAME }}
           source-sha: ${{ inputs.sha || github.sha }}
       - name: Upload additional artifacts
--- a/.github/workflows/conda-cpp-build.yaml
+++ b/.github/workflows/conda-cpp-build.yaml
@@ -2,7 +2,7 @@ on:
   workflow_call:
     inputs:
       build_type:
-        description: "One of: [branch, nightly, pull-request]"
+        description: "One of: [branch, nightly, pull-request, release-candidate]"
         required: true
         type: string
       branch:
@@ -44,6 +44,10 @@ on:
         default: true
         required: false
         description: "One of [true, false], true if artifacts should be uploaded to GitHub's artifact store"
+      candidate-train-sha256:
+        description: "Canonical SHA-256 of the release train; required for release-candidate builds."
+        type: string
+        default: ""
       matrix_filter:
         description: |
           jq expression which modifies the matrix.
@@ -200,7 +204,7 @@ jobs:
           MAMBA_USE_SHARDED_REPODATA: false
           RATTLER_SHARDED: false
       - name: Get Package Name and Location
-        if: ${{ inputs.upload-artifacts }}
+        if: ${{ inputs.upload-artifacts || inputs.build_type == 'release-candidate' }}
         env:
           # Pass RAPIDS_PACKAGE_NAME from cpp-build step if available
           RAPIDS_PACKAGE_NAME: ${{ steps.cpp-build.outputs.rapids-package-name }}
@@ -214,22 +218,22 @@ jobs:
           echo "CONDA_OUTPUT_DIR=${RAPIDS_CONDA_BLD_OUTPUT_DIR}" >> "${GITHUB_OUTPUT}"
         id: package-name
       - name: Show files to be uploaded
-        if: ${{ inputs.upload-artifacts }}
+        if: ${{ inputs.upload-artifacts || inputs.build_type == 'release-candidate' }}
         env:
           CONDA_OUTPUT_DIR: ${{ steps.package-name.outputs.CONDA_OUTPUT_DIR }}
         run: |
           echo "Contents of directory to be uploaded:"
           ls -R "${CONDA_OUTPUT_DIR}"
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
-        if: ${{ inputs.upload-artifacts }}
+        if: ${{ inputs.upload-artifacts && inputs.build_type != 'release-candidate' }}
         with:
           if-no-files-found: 'error'
           name: ${{ steps.package-name.outputs.RAPIDS_PACKAGE_NAME }}
           path: ${{ steps.package-name.outputs.CONDA_OUTPUT_DIR }}
 
       - name: Create Conda release catalog companion
-        if: ${{ inputs.upload-artifacts }}
-        uses: rapidsai/shared-actions/release-catalog-dispatch@32a308ab234f7381dd611056747b4c1ac30d2a34 # shared-actions PR 136 canary head
+        if: ${{ inputs.build_type == 'release-candidate' }}
+        uses: rapidsai/shared-actions/release-catalog-dispatch@8fef990732014851c3d5afcf931c2506f7e175b0 # shared-actions PR 136 candidate-store head
         with:
           config: >-
             {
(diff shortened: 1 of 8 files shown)


.github/workflows/conda-cpp-build.yaml
```diff
@@ -44,6 +44,14 @@ on:
...
         required: false
         description: "One of [true, false], true if artifacts should be uploaded to GitHub's artifact store"
+      release-unit:
+        type: string
+        default: ''
+        required: false
+        description: >-
+          Release component ID: a string label, not a file or bundle. It is written to each manifest entry and groups
+          files and matrix variants for release assembly. Leave empty to use conda:<repository-name>; override only
+          when one repository's Conda producers must be tracked as distinct components.
```
jameslamb: > _override only when one repository's Conda produces must be tracked as distinct components_

I don't understand what this means.

Offline you mentioned testing with these PRs:

* https://github.com/rapidsai/dask-cuda/pull/1672
* https://github.com/rapidsai/nx-cugraph/pull/274
* https://github.com/NVIDIA/cuvs/pull/2400

None of them appear to set `release-unit`... is there a specific repo you're targeting where you'd envision this being used? A concrete example might help me understand why this is needed.

If you don't have a specific example where this would be applicable, let's remove it and simplify here a bit.

I'm also wondering ... wouldn't `conda:<repository-name>` conflict if you have `conda-cpp-build.yaml` and `conda-python-build.yaml` workflow calls in the same CI workflow? Or multiple `conda-python-build.yaml` workflow calls like we do in `cudf` separating noarch and other builds? ([cudf code link](https://github.com/NVIDIA/cudf/blob/a1dc06841c71fd22ef141c8183c57500def71b0f/.github/workflows/pr.yaml#L23-L24))

.github/workflows/custom-job.yaml
```diff
@@ -105,6 +105,15 @@ on:
         default: false
         type: boolean
         required: false
+      release-build-output:
+        description: >-
+          Optional JSON object describing a custom release package bundle. A non-empty value creates the companion;
+          leave empty for logs, documentation, tests, and other non-release artifacts. Requires artifact_type set to
+          custom, component_id, artifacts, and exactly one of package or package_file; output_directory defaults to
+          '.'. The shared action validates every field before inspecting build outputs.
```
jameslamb:
```suggestion
      release-build-config:
        description: >-
          Optional JSON string with additional configuration for the `shared-actions/release-build-output-dispatch` action. See https://github.com/rapidsai/shared-actions/tree/main/release-build-output for details. If non-empty, creates a release manifest for artifacts produced by this job.
```

This is passed to the `config:` input of `shared-actions/release-build-output-dispatch`.

Given that:

1. I think `release-build-output` is a confusing name and something that clarifies it's a configuration would be better.
2. The description should clarify that + redirect to the actions' docs/code for how to find the current schema, instead of literally describing some of the schema and implementation (likely to become out of date as the action changes over in `shared-actions`).
3. The docs should make it much clearer that setting a non-empty value leads to uploading new data. The LLMism "A non-empty value creates the companion" was not clear to me until I read the code and saw `if: ${{ inputs.release-build-output != '' }}`.

My inline suggestion here covers all these points, feel free to adjust as you'd like.

README.md
```diff
@@ -20,6 +20,88 @@ Reusable workflows must be placed in the `.github/workflows` directory as mentio
 
 ## Usage
 
+### release-build-output
```
jameslamb: I think this entire section's contents should be deleted and replaced with a simpler rewording.

I think this could be significantly simplified without any information loss, and that most of the details about how this works should be deferred to a README in `shared-actions` close to the relevant code (not here in ` shared-workflows`).

Some details here also appear to be the output of an LLM responding to a prompt to generate the docs, not information that'd help users of `shared-workflows`. For example, this doesn't belong in the top-level README of the `shared-workflows` repo:

> _The cross-repository enrollment inventory, blockers, and proposed PR sequence
are maintained in
[`rapidsai/build-infra#381`](https://github.com/rapidsai/build-infra/issues/381)._

Suggesting a rewording here, mainly to help others reading these docs but also as a way to test that I've understood this functionality correctly.

----

## (James's proposed rewording below)

Workflows that produce packages intended for redistribution (conda packages, wheels, tarballs, etc.) also upload files containing metadata about the workflows that produced those packages, licenses for source code used in the packages, and more.

Those files are bundled in an archive that roughly looks like this:

```text
.
├── release-build-metadata.json
├── release-build-output.json
└── release-evidence
    ├── <package ID>.provenance.json
    └── <package ID>.spdx.json
```

Those are used at release time to create SBOMs and other provenance records attesting to where/when packages were built and what they contain.

For standardized package types like conda packages and wheels, the naming conventions and process for generating these contents is completely automated.

For custom package types, the `custom-job.yaml` workflow allows specification of a configuration for additional data to be appended. For example, configuration similar to this:

```yaml
cuvs-java-build:
  uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@main
  with:
    # existing build inputs omitted
    artifact-name: cuvs-java-cuda12.9.1
    file_to_upload: java/cuvs-java/target/
    release-build-output: >-
      {
        "artifact_type": "custom",
        "component_id": "maven:cuvs-java",
        "output_directory": "java/cuvs-java/target",
        "package_file": "cuvs-java.release-package.json",
        "artifacts": [{"path": "cuvs-java-*-x86_64-cuda*.jar"}]
      }
```

Results in the same directory layout as shown above, but with content like the following in `release-build-output.json`:

```json
{
  "artifacts": [
    {
      "package": {
        "ecosystem": "maven",
        "name": "com.nvidia.cuvs:cuvs-java",
        "version": "26.10.0"
      },
      "path": "cuvs-java-26.10.0-x86_64-cuda13.jar",
      "provenance": "release-evidence/cuvs-java-26.10.0-x86_64-cuda13.jar.e47b54cfc174f4c5bf3054d99e1f595d26a1ba9ece2cba7c7ba5aba292765892.provenance.json",
      "sbom": "release-evidence/cuvs-java-26.10.0-x86_64-cuda13.jar.e47b54cfc174f4c5bf3054d99e1f595d26a1ba9ece2cba7c7ba5aba292765892.spdx.json",
      "unit_id": "maven:cuvs-java"
    }
  ],
  "producer": "release-platform",
  "schema_version": 1
}
```

For much more detail on how to configure and use these outputs, see `<link to docs in 'shared-actions' repo>`.

.github/workflows/custom-job.yaml
```diff
@@ -228,6 +237,13 @@ jobs:
           name: ${{ inputs.artifact-name }}
           path: ${{ inputs.file_to_upload }}
           if-no-files-found: ignore
+      - name: Create release build-output companion
+        if: ${{ inputs.release-build-output != '' }}
+        uses: rapidsai/shared-actions/release-build-output-dispatch@a18a4a7ac572366c09c15641ec274cc6f15bfb5d # agent/release-build-output-container
+        with:
+          config: ${{ inputs.release-build-output }}
+          source-artifact-name: ${{ inputs.artifact-name }}
+          source-sha: ${{ inputs.sha || github.sha }}
```
jameslamb: There are multiple other uses of `inputs.sha` above. What's a case where falling back to `{{ github.sha }}` would be appropriate / expected here?

If you don't have a specific case for that, recommend using `{{ inputs.sha }}`.

(similar question applies to all cases like this in this PR)

jameslamb (dismissed):
Tried my best to review. I have enough questions about this that I'm leaving a blocking review.

.github/workflows/conda-cpp-build.yaml
msarahan: I struggled with the "release unit" name as well. Believe it or not, it had a worse name before. Sorry I didn't get this to be more understandable.

A release-unit is a collection of files that shares:
* validation/posting state
* whether it should be included in the release (perhaps useful for something like your selective deployment of UCXX hotfix release)
* dependencies (within our graph, not something like python dependencies)
* destination (so you could split normal wheels from wheels that only go to pypi.nvidia.com, for example)

A given filename within a release unit must be unique, but multiple release units with the same name can/will be merged across multiple manifest files.

We can get rid of this for simplicity. I don't think we have a use case for it currently. I think it is worth keeping the definition of what a release unit is, though.

.github/workflows/custom-job.yaml
msarahan: Big improvement. Thank you.

.github/workflows/custom-job.yaml
msarahan: Makes sense. The LLM wanted to fall back to env.RAPIDS_SHA, which is apparently set in each of the workflows in one of two ways:

* Conda C++ directly writes git rev-parse HEAD to RAPIDS_SHA: [conda-cpp-build.yaml (line 138)](/Users/msarahan/code/.codex-worktrees/shared-workflows-release-build-output-dispatch/.github/workflows/conda-cpp-build.yaml:138) 
* Conda Python does the same: [conda-python-build.yaml (line 148)](/Users/msarahan/code/.codex-worktrees/shared-workflows-release-build-output-dispatch/.github/workflows/conda-python-build.yaml:148)
* Wheels invokes rapids-github-info: [wheels-build.yaml (line 176)](/Users/msarahan/code/.codex-worktrees/shared-workflows-release-build-output-dispatch/.github/workflows/wheels-build.yaml:176)
* custom-job invokes it too: [custom-job.yaml (line 180)](/Users/msarahan/code/.codex-worktrees/shared-workflows-release-build-output-dispatch/.github/workflows/custom-job.yaml:180)

I think it's fine to either not be defensive with the fallback, or to error out when it is not set.

EDIT: several workflows do not pass this, and instead rely on the state of the already checked-out source. env.RAPIDS_SHA seems like the right way. +1 for the LLM over me.

msarahan:
Marking this as do not merge because we should change the shared-actions refs after merging https://github.com/rapidsai/shared-actions/pull/136

PR #610, opened 2026-08-03
Disable GB300 for PRs

KyleFromNVIDIA:


bdice (approved):
Thanks. We only have enough gb300 capacity for nightlies right now.

PR #614, opened 2026-08-04
X-ORG-596: Only publish numeric version of docs

josephine-wolf-oberholtzer:
Contributes to #596 

For the shared workflow, let's only push the XX.YY versions of the docs to docs.nvidia.com. Pushing the friendly tags ("stable", "legacy", "nightly") is tricky to get correct when performed from a per-repo, per-branch build, so we're going to punt on that technique for this release and let it cook more.

Instead we'll use the deploy workflow (https://github.com/rapidsai/docs/pull/820) to guarantee the correct friendly tags ("stable", "legacy", "nightly") are pushed to docs.nvidia.com in a coordinated way on a daily / on-demand basis, just like for docs.rapidsai.org.

To simplify backwards-compatibility, and to leave the door open to coordinating friendly tags on a per-repo basis, I'm leaving the "version-map" input as-is. We're just going to stop using it for the time being.

### DIFF
--- a/.github/workflows/publish-api-docs.yaml
+++ b/.github/workflows/publish-api-docs.yaml
@@ -110,15 +110,6 @@ jobs:
         run: |
           echo rapids-version-major-minor="$(rapids-version-major-minor)" >> "${GITHUB_OUTPUT}"
 
-      - name: Resolve VERSION tag
-        id: version-tag
-        env:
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
Loving that negative diff

But yeah, I like this plan

PR #617, opened 2026-08-20
Change pr-builder runner from ubuntu-latest to ubuntu-slim

bdice:
GitHub’s `ubuntu-latest` runners are heavily backlogged, and `pr-builder` cannot run, which blocks merges. Switching to use self-hosted CPU runners should help alleviate the problem.

### DIFF
--- a/.github/workflows/pr-builder.yaml
+++ b/.github/workflows/pr-builder.yaml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   run:
-    runs-on: linux-amd64-cpu4
+    runs-on: ubuntu-slim
     steps:
       # This reusable workflow should depend on all other jobs in the calling workflow.
       - name: "Check all dependent jobs"
--- a/.github/workflows/pr-builder.yaml
+++ b/.github/workflows/pr-builder.yaml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   run:
-    runs-on: ubuntu-latest
+    runs-on: linux-amd64-cpu4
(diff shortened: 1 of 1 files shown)


gforsyth:
Do we still want to do this?  

@jakirkham mentioned we've had our limit bumped for those runners last night at ~9pm eastern: https://nvidia.slack.com/archives/C01PF84NXCL/p1787186487740469?thread_ts=1787181186.430319&cid=C01PF84NXCL

jameslamb:
I just asked our GitHub contacts a clarifying question... `pr-builder` might be a good fit for `ubuntu-slim` runners: https://github.blog/changelog/2025-10-28-1-vcpu-linux-runner-now-available-in-github-actions-in-public-preview/

Those are physically smaller and we might face higher concurrency limits for them... if so I'd rather switch to that than put more burden on our self-hosted CPU runners.

bdice:
The reason I chose self-hosted infrastructure is because it isolates us from the recurring issues in GitHub infrastructure. I'll try `ubuntu-slim`, please approve and merge if you are fine with that.

.github/workflows/pr-builder.yaml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   run:
-    runs-on: ubuntu-latest
+    runs-on: linux-amd64-cpu4
```
bdice:
```suggestion
    runs-on: ubuntu-slim
```

PR #618, opened 2026-08-24
Add continue-on-error option to conda CPP and Python tests

wence-:
The wheels-test shared workflow already offers this option. Let's do so for the conda test workflows too.


jameslamb (approved):
You're right. We did this in https://github.com/rapidsai/shared-workflows/pull/444, it's fine to do here for conda too.

wence-:
Thanks!

